PR #8, opened 2015-10-12
defcon: Add TGlyph.autoUnicode() to get unicode from AGL or uninames

moyogo:


adrientetar:
Thanks. In the future we'll make it possible to change the AGL for another glyphList so people can input their development names.

moyogo:
> In the future we'll make it possible to change the AGL for another glyphList so people can input their development names.

Maybe something like GlyphOrderAndAliasDB, that can also be imported or exported?

adrientetar:
Yes.

Issue #7, opened 2015-10-12
Deleting points in OSx


pathumego:
In OSx 10.10.5, To delete points I have to hold `fn+Delete` Expected key is just Delete.

davelab6:
> Expected key is just Delete

ie, backspace

diegommaldo:
Beyond that, if I try to delete off curve points, in a broken angle bezier, when I delete one, it deletes the the other instead. When the bezier off curve is aligned, works fine.

Mac OS 10.11

moyogo:
There is no ⌦ Delete key on Mac Book keyboards, fn + ← Backspace is what you have to use for the Delete key. The ⌦ Delete key is on 105-key keyboards next to ↖ Home, ↘ End, ⭻ Page Up, ⭽ Page Down.
@pathumego Do you want Backspace to behave like the Delete key and delete selected nodes, etc.?

> Beyond that, if I try to delete off curve points, in a broken angle bezier, when I delete one, it deletes the the other instead. When the bezier off curve is aligned, works fine.

@diegommaldo This sound like a separate issue.

diegommaldo:
for sure it is, but we are talking about deleting points, right @moyogo ?

adrientetar:
> in a broken angle bezier

I do not understand this.

pathumego:
:+1:

PR #12, opened 2015-10-14
Make Backspace key work as Delete key

moyogo:
See #7

### DIFF
--- a/Lib/defconQt/glyphCollectionView.py
+++ b/Lib/defconQt/glyphCollectionView.py
@@ -228,7 +228,7 @@ class GlyphCollectionWidget(QWidget):
             self.selection = set()
         # XXX: this is specific to fontView so should be done thru subclassing of a base widget,
         # as is done in groupsView
-        elif key == Qt.Key_Delete:
+        elif key in (Qt.Key_Delete, Qt.Key_Backspace):
             #if self.characterDeletionCallback is not None:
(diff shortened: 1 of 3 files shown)


Lib/defconQt/glyphCollectionView.py
```diff
@@ -228,9 +228,9 @@ def keyPressEvent(self, event):
             self.selection = set()
         # XXX: this is specific to fontView so should be done thru subclassing of a base widget,
         # as is done in groupsView
-        elif key == Qt.Key_Delete:
+        elif key in (Qt.Key_Delete, Qt.Key_Backspace):
             #if self.characterDeletionCallback is not None:
-            if proceedWithDeletion() and self.selection:
+            if proceedWithDeletion(self) and self.selection:
```
adrientetar: Remove self in the function definition instead.

adrientetar:
It should dispatch according to the platform, Backspace isn't usually used on Windows for this kind of operation.

adrientetar:
I'll merge this manually.

moyogo:
There are probably other behaviours that need be different according to the platform. How do you want to handle these?

adrientetar:
Okay, I think I have sorted this out in master. Also, your earlier commit was the good one so I recovered it.

moyogo:
Thanks!

PR #17, opened 2015-10-15
Layer features

graphicore:
Ok, this is a rather big change, so you should check it out and see if you like it.

Also, I had two bugfixes in my defcon branch. You may want to update.

I started at one end and worked me through a bulk of stuff, I hope the changes are not to much for your taste.

I changed a lot how GlyphView is organzied in a) terms of how it draws and redraws b) I changed some things regarding object hierarchy. MainGfxWindow now acts much more as a controller for GlyphView. For example, when the glyph is changed MainGfxWindow swaps out the GlyphView object. Some control of the layerSet is happening in MainGfxWindow etc.

What I achieved is almost feature parity with defcon/ufo3 layers, it's shorter to keep out what we don't do:

* delete the default layer: I feared that rest of the app expects a default layer to be present
* change the default layer: should be easier to do, I just thought it's not that important at the moment

what we do:
* add
* delete
* rename
* change Color
* reorder

All layer-tools a properly connected to layer/layerSet events, so you can change the layer setup in the new FontWindow -> Windows -> LayerSet window and your opened glyph window will react and vice versa.

You can also work on a layer in one window and see the changes propagated in another window, even if the layer is not the active one. Stuff like this does not work in master.

I added a "Display" menu to the glyph view with these options:

* Active Layer Filled
* Active Layer use Custom Color
* Other Layers Filled
* Other Layers use Custom Color
* Show Other Layers

(we can easily add more)

and two presets, that act upon the previous options:

* *TruFont Default*: looks like the current deployed version
* *Layer Fonts*: fills all layers, no outlines, layers are in layerOrder

The second mode is suitable if you want to design layer fonts, like these: https://www.google.de/search?q=layer+fonts

selection_080


adrientetar:
<3<3!

I'll take a look.

moyogo:
Nice!

Wouldn’t it be better to have these layer display settings as icon-like checkboxes in the layer window next to the laye name and color?

graphicore:
I'm sure there's a lot room for UI/UX improvements :-)

Lib/defconQt/glyphView.py
```diff
@@ -1,14 +1,16 @@
 from enum import Enum
 from math import copysign
+from functools import partial
 import pickle
 from defcon import Anchor, Component
 from defconQt import icons_db
 from defconQt.objects.defcon import TContour, TGlyph
 from defconQt.pens.copySelectionPen import CopySelectionPen
 from defconQt.util import platformSpecific
+from defconQt import icons_db
```
adrientetar: Have you rebased on top of master? It seems you merged the latest changes in the tree in your commits.

Lib/defconQt/glyphView.py
```diff
@@ -187,27 +292,38 @@ def __init__(self, glyph, parent=None):
...
+        selectionToolButton = toolBar.addAction("Selection", self._redirect('view', 'setSceneSelection'))
         selectionToolButton.setCheckable(True)
         selectionToolButton.setChecked(True)
         selectionToolButton.setIcon(QIcon(":/resources/cursor.svg"))
-        penToolButton = toolBar.addAction("Pen", self.view.setSceneDrawing)
+        penToolButton = toolBar.addAction("Pen", self._redirect('view', 'setSceneDrawing'))
         penToolButton.setCheckable(True)
-        penToolButton.setIcon(QIcon(":/resources/curve.svg"))
-        rulerToolButton = toolBar.addAction("Ruler", self.view.setSceneRuler)
+        selectionToolButton.setIcon(QIcon(":/resources/cursor.svg"))
```
adrientetar: Why this change?

Lib/defconQt/glyphView.py
```diff
@@ -187,27 +292,38 @@ def __init__(self, glyph, parent=None):
...
+        rulerToolButton = toolBar.addAction("Ruler", self._redirect('view', 'setSceneRuler'))
```
adrientetar: Make the toolbar (and all UI elements controlled by the view) be created in the View and returned with a getter function.  
e.g. see what I did with `view._currentLayerBox` in the View, then the MainGfxWindows calls view.currentLayerBox() to get the widget and put it in the layout.

This should allow removing the _redirect function.

Lib/defconQt/glyphView.py
```diff
@@ -187,27 +292,38 @@ def __init__(self, glyph, parent=None):
...
+        knifeToolButton = toolBar.addAction("Knife", self._redirect('view', 'setSceneKnife'))
         knifeToolButton.setCheckable(True)
-        knifeToolButton.setIcon(QIcon(":/resources/cut.svg"))
+        selectionToolButton.setIcon(QIcon(":/resources/cursor.svg"))
         # http://www.setnode.com/blog/right-aligning-a-button-in-a-qtoolbar/
         spacer = QWidget()
         spacer.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
         toolBar.addWidget(spacer)
-        toolBar.addWidget(self.view.currentLayerBox())
+        self._currentLayerBox = QComboBox(self)
```
adrientetar: Why?

adrientetar: Is the goal to have all the external UI elements in the MainWindow rather than in the View? I've found handling that in the View more convenient in practise.

Lib/defconQt/glyphView.py
adrientetar: Okay it seems you did this purposefully, what's the rationale for the change?

Lib/defconQt/glyphView.py
```diff
@@ -217,34 +333,200 @@ def __init__(self, glyph, parent=None):
...
+            oldView.hide()
+            oldView.deleteLater()
+
+        self.view = GlyphView(glyph, self._displaySettings, self)
+        # Preserve the zoom level of the oldView.
+        # FIXME: is there more state that we need to carry over to the next
+        # GlyphView? (This seems pretty complete to me…) If yes, we should
+        # make an interface for a predecessor -> successor transformation
+        if oldView:
+            self.view.setTransform(oldView.transform())
```
adrientetar: I feel that e.g. what happens in this function is making things more complex and less straightforward to reason about. What is the benefit for this downside?

Why creating a new GlyphView rather than redrawing things inside of it?

Lib/defconQt/glyphView.py
```diff
@@ -217,34 +333,200 @@ def __init__(self, glyph, parent=None):
...
+        self._setlayerColorButtonColor()
+
+        app = QApplication.instance()
+        app.setCurrentGlyph(glyph)
+        self.setWindowTitle(glyph.name, glyph.getParent())
+
+        # switch
+        self.setCentralWidget(self.view)
+
+    def _excuteRemoteCommand(self, targetName, commandName, *args, **kwds):
```
adrientetar: typo: execute

Lib/defconQt/glyphView.py
```diff
@@ -217,34 +333,200 @@ def __init__(self, glyph, parent=None):
...
+        target = getattr(self, targetName)
+        command = getattr(target, commandName)
+        command(*args, **kwds)
+
+    def _redirect(self, target, command):
+        """
+        This creates an indirection to permanently connect an event to a
+        property that may (will) change its identity.
+        """
+        return partial(self._excuteRemoteCommand, target, command);
```
adrientetar: I'm not sure I understand what this layer of indirection brings us.

Lib/defconQt/glyphView.py
```diff
@@ -217,34 +333,200 @@ def __init__(self, glyph, parent=None):
...
+        else:
+            layer.addObserver(self, '_layerColorChange', 'Layer.ColorChanged')
+
+    def _chooseLayerColor(self):
+        if QApplication.keyboardModifiers() & Qt.ShiftModifier:
+            # reset to default color, i.e. delete the layer color
+            color = None
+        else:
+            startColor = self.view.getLayerColor() or QColor('limegreen')
+            qcolor = QColorDialog.getColor(startColor, self, options=QColorDialog.ShowAlphaChannel)
```
adrientetar: qColor

Lib/defconQt/glyphView.py
```diff
@@ -217,34 +333,200 @@ def __init__(self, glyph, parent=None):
...
+
+    def _displayChanged(self, *args):
+        # redraw the view
+        self.view.redraw()
 
     def event(self, event):
         if event.type() == QEvent.WindowActivate:
             app = QApplication.instance()
+            # FIXME! which layer is this then -- default?
             app.setCurrentGlyph(self.view._glyph)
```
adrientetar: Regarding your FIXME, a glyph always belongs to just one layer, so shouldn't view._glyph host the glyph from the current layer?

Lib/defconQt/glyphView.py
```diff
@@ -217,34 +333,200 @@ def __init__(self, glyph, parent=None):
...
         return super(MainGfxWindow, self).event(event)
 
     def closeEvent(self, event):
-        self.view._glyph.removeObserver(self, "Glyph.Changed")
-        event.accept()
+        # FIXME: somehow the reference to this window is not deleted from
+        # memory, so the weakrefs of defcon notifications stay in tact.
+        # Must do this manually :-(
```
adrientetar: It's explicitly said in defcon documentation that you have to remove observers yourself or it will keep sending notifications into the closed widget.

Lib/defconQt/glyphView.py
```diff
@@ -677,7 +959,6 @@ def __init__(self, path, component, parent=None):
         super(ComponentItem, self).__init__(path, parent)
         self._component = component
         self.setTransform(QTransform(*component.transformation))
-        self.setBrush(QBrush(componentFillColor))
```
adrientetar: Why this change?

Lib/defconQt/glyphView.py
```diff
@@ -1341,17 +1633,30 @@ def knifeMouseRelease(self, event):
...
+
+        # will change during lifetime
+        self._layer = glyph.layer
+
         self._impliedPointSize = 1000
         self._pointSize = None
 
+        # Where is this stuff used?
         self._inverseScale = 0.1
         self._scale = 10
```
adrientetar: It's not, there's some dead code in there.

Lib/defconQt/glyphView.py
```diff
@@ -1359,15 +1664,12 @@ def __init__(self, glyph, parent=None):
         self._showMetricsTitles = True
 
         self.setBackgroundBrush(QBrush(Qt.lightGray))
-        self.setScene(GlyphScene(self))
+        self.setScene(GlyphScene(self, self._getSceneItems('scene-added-items')))
```
adrientetar: Okay, we're taking objects and carrying them to a new Scene... isn't a setGlyph approach simpler?

Lib/defconQt/glyphView.py
graphicore: When a glyph is changed I create a new GlyphView and throw away the old one. Stuff that I want to keep is handled by the Window. I do this, so I don't need to keep track of state changes that happen between glyph changes (like subscriptions to the glyph object), I think switching layers etc. whatnot is enough to manage in GlyphView

the switch is performed here: https://github.com/graphicore/trufont/commit/d2a3e6dcc58661ff6751af5f9906a27c9df4f5e9#diff-4afada8b45ba97a28f2ff0c5a9995dc0R351



That's what I meant with:

> I changed some things regarding object hierarchy. MainGfxWindow now acts much more as a controller for GlyphView.

The redirect thing keeps me from having to rewire the signals after I replaced the view.

Lib/defconQt/glyphView.py
```diff
@@ -1380,68 +1682,127 @@ def __init__(self, glyph, parent=None):
...
         item = scene.addRect(-1000, -1000, 3000, 3000, QPen(Qt.black), QBrush(Qt.gray))
         item.setZValue(-1000)
         scene._widthItem = scene.addRect(0, -1000, width, 3000, QPen(Qt.NoPen), QBrush(backgroundColor))
         scene._widthItem.setZValue(-999)
-        descender = font.info.descender
-        if descender is None: descender = -250
-        unitsPerEm = font.info.unitsPerEm
-        if unitsPerEm is None: unitsPerEm = 1000
-        self.centerOn(width/2, descender+unitsPerEm/2)
+        self.centerOn(width/2, font.info.descender+font.info.unitsPerEm/2)
```
adrientetar: What if font.info attributes are None? Won't this crash?

Lib/defconQt/glyphView.py
```diff
@@ -1472,7 +1833,7 @@ def addHorizontalMetrics(self):
             ("Baseline", 0),
             ("x-height", font.info.xHeight),
             ("Cap height", font.info.capHeight),
-            ("Ascender", font.info.ascender),
+            ("Ascender", font.info.ascender)
```
adrientetar: I like trailing comma as it spares you two lines of diff when you add an attribute.

Lib/defconQt/glyphView.py
```diff
@@ -1500,56 +1864,178 @@ def addHorizontalMetrics(self):
...
             item = AnchorItem(anchor, self.transform().m11())
-            item.setZValue(-996)
+            item.setZValue(-992)
+            anchors.append(item)
             scene.addItem(item)
 
     def addPoints(self):
         scene = self.scene()
+
+        ocPoints = self._getSceneItems('ocPoints', clear=True)
```
adrientetar: oc doesn't really say whether its onCurves or offCurves... also it would be oCPoints but more descriptive name doesn't hurt

Lib/defconQt/glyphView.py
```diff
@@ -1651,62 +2138,52 @@ def createAnchor(self):
...
-
-    def setGlyph(self, glyph):
+    def _makeLayerGlyph(self, layer):
+        name = self._name
+        layer.newGlyph(name)
+        glyph = layer[name]
+        # TODO: generalize this out, can’t use newStandardGlyph unfortunately
+        glyph.width = self.defaultWidth
+
+        # What does this do actually?
```
adrientetar: Sets the glyph as template glyph, i.e. a shallow glyph that is displayed in fontView but not saved in the font, only in the glyphOrder. Useful to make and display the initial glyphSet or your font.

Lib/defconQt/glyphView.py
```diff
@@ -1651,62 +2138,52 @@ def createAnchor(self):
...
         super(GlyphView, self).showEvent(event)
         font = self._glyph.getParent()
-        # TODO: we should have an app-wide mechanism to handle default metrics
-        # values (that are applied to new fonts as well)
-        descender = font.info.descender
-        if descender is None: descender = -250
-        unitsPerEm = font.info.unitsPerEm
-        if unitsPerEm is None: unitsPerEm = 1000
-        self.fitInView(0, descender, self._glyph.width, unitsPerEm, Qt.KeepAspectRatio)
+        self.fitInView(0, font.info.descender, self._glyph.width, font.info.unitsPerEm, Qt.KeepAspectRatio)
```
adrientetar: same as before about font.info

Lib/defconQt/glyphView.py
graphicore: it does

adrientetar:
That should be it for code, thanks. Next up I'll make UI comments.

Lib/defconQt/glyphView.py
graphicore: where? and why is it using weakref then? https://github.com/graphicore/defcon/blob/python3/Lib/defcon/tools/notifications.py#L189

Lib/defconQt/glyphView.py
graphicore: it's done in addComponent now, so that the component can be styled from a common method

Lib/defconQt/glyphView.py
adrientetar: I haven't looked in depth at this code, just that the documentation says 
https://github.com/graphicore/defcon/blob/python3/documentation/source/concepts/notifications.rst#dont-forget-removeobserver

Lib/defconQt/glyphView.py
graphicore: I'm not  a huge friend of objects down in the hierarchy using object interfaces further on top. It's bad for refactoring and portability. Thus, you inject them with explicit interfaces.

This is used to delete the items the scene adds, it's a reference to the "registry" list. The scene only adds items to the list, it does not use them.

see this comment of yours: https://github.com/trufont/trufont/blob/master/Lib/defconQt/glyphView.py#L1426

> ```
            # XXX: discriminate better
            if isinstance(item, QGraphicsPathItem) and item.zValue() == -997:
                scene.removeItem(item)
```

File: Lib/defconQt/glyphView.py (hunk 9 above)
Comment (adrientetar) [inline Lib/defconQt/glyphView.py]: So why the FIXME?

File: Lib/defconQt/glyphView.py (hunk 13 above)
Comment (adrientetar) [inline Lib/defconQt/glyphView.py]: Okay.

File: Lib/defconQt/glyphView.py (hunk 14 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: right, when I did this it was pre:

https://github.com/trufont/trufont/commit/a80988874c5d621a2b11c75f5af31b23d3008ee7#diff-4afada8b45ba97a28f2ff0c5a9995dc0L1420

Sorry for this. I did some hand merging and missed this :-/

File: Lib/defconQt/glyphView.py (hunk 15 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: Yeah, NP. In JavaScript it's illegal, and I'm writing a lot of that ...
I use a lot of comma first style for the same reason though.

File: Lib/defconQt/glyphView.py (hunk 17 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: Yeah, but layer glyphs get saved actually.

File: Lib/defconQt/glyphView.py (hunk 18 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: same reason as well, will fix: https://github.com/trufont/trufont/commit/a80988874c5d621a2b11c75f5af31b23d3008ee7#diff-4afada8b45ba97a28f2ff0c5a9995dc0L1420

File: Lib/defconQt/glyphView.py (hunk 4 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: So, the active layer is kept in the window when glyph is changed (the GlyphView is deleted and replaced by a new one)

> Is the goal to have all the external UI elements in the MainWindow rather than in the View? I've found handling that in the View more convenient in practise.

I wouldn't call it the goal. I rather wanted to create a kind of a hierarchy. The layers as they are implemented in ufo/defcon come with this strange concept, that each layer is in a way a font by it's own. To reflect that structure in ufo defcon has: font->layerSet->layers->glyphs and not, as one would intuitively think: font->glyphs->layers

I think it's good not to have a too big gap between the model (defcon) hierarchy and the view (UI) hierarchy. There's kind of a "natural" place for things to happen. Some things happen in the glyph, and some happen one level above.

File: Lib/defconQt/glyphView.py (hunk 5 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: So that we can just throw away state that we don't need and don't want to manage.

File: Lib/defconQt/glyphView.py (hunk 17 above)
Comment (adrientetar) [inline Lib/defconQt/glyphView.py]: Hm, I think it's because the save exclusion mechanism wasn't upgraded to the ufo3 method as in https://github.com/typesupply/defcon/issues/24#issuecomment-144384072

File: Lib/defconQt/glyphView.py (hunk 18 above)
Comment (adrientetar) [inline Lib/defconQt/glyphView.py]: Okay. Creating a new branch and cherry-picking commits is the best way to rebase in my opinion.

File: Lib/defconQt/glyphView.py (hunk 9 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: I think the question is rather whether the global application should receive the default glyph or the real current glyph element. I think it's good like it is, because there may not always be a default glyph. I will remove the comment.

File: Lib/defconQt/glyphView.py (hunk 10 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: Let's figure out what to do there: https://github.com/typesupply/defcon/issues/12#issuecomment-148487917

File: Lib/defconQt/glyphView.py (hunk 17 above)
Comment (graphicore) [inline Lib/defconQt/glyphView.py]: ok, so I'll remove the template marker, because we actually want to keep the glyph? or should we remove the marker when the glyph is changed first?

Comment (adrientetar) [inline Lib/defconQt/glyphView.py]: Yes in the custom defcon class when glyph.dirty is set glyph.template is unset.

Comment (graphicore) [inline Lib/defconQt/glyphView.py]: Then I just remove the comment.

**Comment by graphicore** [comment]
@moyogo 

> Wouldn’t it be better to have these layer display settings as icon-like checkboxes in the layer window next to the laye name and color?

If we do so, it would also be nice to persist that setting in the ufo, which is not in the standard but a good fit for the `lib` of layerinfo plist. I think I was just looking to support the already existing layers operations. I wouldn't add it to this PR.

**Comment by graphicore** [comment]
@adrientetar I updated the PR and will probably not touch it until further conversation indicates.

**Comment by moyogo** [comment]
@graphicore 
> I wouldn't add it to this PR.

You’re right.

File: Lib/defconQt/glyphView.py [hunk 19]
```diff
@@ -217,21 +333,177 @@ def __init__(self, glyph, parent=None):
...
+        self._setlayerColorButtonColor()
+
+        app = QApplication.instance()
+        app.setCurrentGlyph(glyph)
+        self.setWindowTitle(glyph.name, glyph.getParent())
+
+        # switch
+        self.setCentralWidget(self.view)
+
+    def _exceuteRemoteCommand(self, targetName, commandName, *args, **kwds):
```
adrientetar: still typo: execute

Lib/defconQt/glyphView.py
```diff
@@ -1341,17 +1627,30 @@ def knifeMouseRelease(self, event):
...
-        self._glyph = glyph
-        self._glyph.addObserver(self, "_glyphChanged", "Glyph.Changed")
-        self._glyph.layerSet.addObserver(self, "_layersChanged", "LayerSet.Changed")
+
+        # wont change during lifetime
+        self._layerSet = layerSet = glyph.layerSet
+        self._name = glyph.name
+        self._settings = settings
+
+        # we got the individula layers as keys => pathItem
```
adrientetar: typo

Lib/defconQt/glyphView.py
```diff
@@ -1380,59 +1676,122 @@ def __init__(self, glyph, parent=None):
...
+        self._listenToLayerSet()
+
+        self.changeCurrentLayer(self._layer)
+
+
+    @property
+    def _glyph(self):
+        # instead of a getter we could change the glyph each time
+        # self._layer is changed. However, it might be that the glyph
+        # is not ready when the layer is set
```
adrientetar: What do you mean by not ready?

graphicore: You first create a layer then you can create the glyph. The layer must be present before the glyph.

Lib/defconQt/glyphView.py
graphicore: hm, I thought I got that.

adrientetar:
We won't make a separate layers window. Instead add a panel to the inspector window.

layer

adrientetar:
I have started thinking about how to make panels of inspector window collapsible, I'll work on that afterwards.

adrientetar:
Let's make all the UI self-contained in this panel, except the dropdown which allows to change current layer without opening the inspector window.

graphicore:
You want me to add the layer widget into the inspector window? Can do.

graphicore:
Here you go.

inspector_081

adrientetar:
Nice, just three things:

- Can you LayerSetWidget a subclass of QListWidget directly and call it LayerSetList?
- As for the no layer color image, the one you put is cool and weird. I think, maybe a blank square with a red diagonal trait going top-right to bottom-left would be easier to understand.
- It would be cool to offload the checkboxes in the glyphView menu to buttons above the layers list, but this can be done later.

Lastly, squash your commits and I'll merge this. Thank you!

graphicore:
> Can you LayerSetWidget a subclass of QListWidget directly and call it LayerSetList?

Yes.

> As for the no layer color image, the one you put is cool but odd. I think, maybe a blank square with a red diagonal trait going top-right to bottom-left would be easier to understand.

White background or transparent background?

> It would be cool to offload the checkboxes in the glyphView menu to buttons above the layers list

The checkboxes In the "Display" menu to the layerlist in the Inspector window? This would make them from a local glyphView centric control to a global control for all glyphViews. So you can't have two glyphViews with different looks. May be good or bad, I'm not so deep into UX. Maybe global presets and local overrides would satisfy both models and allow to use this flexible.

The global settings could also be used by the GlyphList in the font window.

> , but this can be done later.

I agree

adrientetar:
> White background or transparent background?

White. This should I think be drawable into a pixmap without external image, though that's a detail.

graphicore:
> This should I think be drawable into a pixmap

Of course it is. But it causes more work for me and so for anyone who wants to change it later (if someone wants)

Lib/defconQt/LayerSetWidget.py
```diff
@@ -0,0 +1,147 @@
+from PyQt5.QtCore import Qt
+from PyQt5.QtGui import QKeySequence, QColor, QPixmap, QIcon
+from PyQt5.QtWidgets import QWidget, QMenu, QListWidget, QListWidgetItem, \
+                            QAbstractItemView, QVBoxLayout, QAction, QColorDialog
+from defconQt import icons_db
+from defconQt.glyphView import AddLayerDialog
+
+from functools import partial
```
adrientetar: (This seems unused in the file.)

graphicore:
Ha, that was a fast response!

Tell me what next or when you are ready for a squashed commit.

adrientetar:
lgtm, I think you can squash now and I'll merge, thanks.

graphicore:
OK

graphicore:
Did you get the message? That last version of the commit includes the change to use `app.currentFont()`. We missed us here. You can either force push it to master, let me create a new commit with just that change plus PR or make it yourself.

Issue #28, opened 2015-10-19
"Smart"-sorting feature


felipesanches:
"Smart sort" does not intuitively describe the feature.

What does it do? We should choose a more descriptive name for this feature.
Behdad suggests avoiding "super"/"Smart"-feature naming scheme in general because it is typically meaningless.
- [ ] Specify the desired feature (for documentation purposes)
- [ ] Choose a better name for it

adrientetar:
> What does it do?

It's called "canned design" in defcon, basically it's a combination of multiple sorts with some rather arbitrary criteria that yield an optimized sorting order that resembles Tal Leming's preferred order.

Documentation says "Canned sorts are combinations of simple, complex and custom sorts that give optimized ordering results."

felipesanches:
Should we call it "Canned sort" as well then? Or even something better?
Maybe we should add tooltip text to these options, so that we can better inform the users.

Anyway... the long description you just provided does not seem much more meaningful to the users than "This feature does a lot of magical sorting stuff" :-)

adrientetar:
> Should we call it "Canned sort" as well then?

Canned is a bit obscure, smart sort says this is an optimized default that was arbitrarily chosen as a good ordering.

> Maybe we should add tooltip text to these options, so that we can better inform the users.

Yes, I will do that with a description.

PR #30, opened 2015-10-19
temporarily remove "smart"-sort from UI as it is not implemented yet.

felipesanches:
This is related to issue #28

### DIFF
--- a/Lib/defconQt/fontView.py
+++ b/Lib/defconQt/fontView.py
@@ -467,7 +467,7 @@ class SortDialog(QDialog):
         self.setWindowModality(Qt.WindowModal)
         self.setWindowTitle("Sort…")
 
-        self.smartSortBox = QRadioButton("Smart sort", self)
+        #self.smartSortBox = QRadioButton("Smart sort", self) #See issue at: https://github.com/trufont/trufont/issues/28
         self.glyphSetBox = QRadioButton("Glyph set", self)
         self.glyphSetBox.toggled.connect(self.glyphSetToggle)
         self.glyphSetDrop = QComboBox(self)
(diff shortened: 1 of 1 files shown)


adrientetar:
It *is* implemented and it works for me. It's the default sort order &ndash; have you applied another sorting order before doing smart sort again?

felipesanches:
All I see is self.smartSortBox being instantiated and set, but it's status does not seem to be used anywhere...

Am I missing something?

Here's what I see:

```
felipe@guarana:~/devel/github_felipesanches/trufont/Lib$ ack-grep smart -C 1
defconQt/fontView.py
469-
470:        self.smartSortBox = QRadioButton("Smart sort", self)
471-        self.glyphSetBox = QRadioButton("Glyph set", self)
--
491-        elif desc[0]["type"] == "cannedDesign":
492:            self.smartSortBox.setChecked(True)
493-        else:
--
535-        layout = QVBoxLayout(self)
536:        layout.addWidget(self.smartSortBox)
537-        layout.addWidget(self.glyphSetBox)
```

adrientetar:
Lines 601&ndash;602:

```
    else:
        ret = cannedDesign
```

felipesanches:
hmm... That was not trivial or evident. Handling the checkbox this way is a bit misleading I think.

adrientetar:
I always make the last else if and else for exhaustive pattern-matching.

felipesanches:
What I mean is that the checkbox state is set by the user, but it is not read by the program. Instead, it's state is infeered by the state of all other checkboxes, which seems bad code style. I'd suggest you actually check the checkbox value in order to decide whether or not the feature will be enabled.

adrientetar:
The checkboxes are mutually-exclusive and there is a default check, so this checkbox being checked is logically equivalent to others being unchecked.

PR #45, opened 2015-10-21
Ubuntu/Debian installation steps + setup.py tweaks

n7s:
Making it a little easier to install and use on Ubuntu/Debian...

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -1,4 +1,50 @@
-To install this package, type the following in the command line:
 
-  python setup.py install
+Installation steps for Ubuntu/Debian
+=====================================
+
+(Note that we are using version 3.4 of python).
+
+
+sudo apt-get install -qq -y python3-pyqt5 python3-flake8
+
+git clone --depth=1 --branch=python3-ufo3 https://github.com/trufont/robofab
+git clone --depth=1 --branch=python3-ufo3 https://github.com/trufont/defcon
+git clone --depth=1 --branch=python3-ufo3 https://github.com/trufont/ufo2fdk
+git clone --depth=1 https://github.com/behdad/fonttools.git
+
+cd defcon
+sudo python3.4 setup.py install --record installed-files.txt
+
+cd ../robofab
+sudo python3.4 setup.py install --record installed-files.txt
+
+cd ../ufo2fdk
+sudo python3.4 setup.py install --record installed-files.txt
+
+cd ../fonttools
+sudo python3.4 setup.py install --record installed-files.txt
+
+cd ../trufont
+sudo python3.4 setup.py install --record installed-files.txt
+
+To run trufont:
+python3.4 -m defconQt
+
+
+To uninstall cleanly the various modules and trufont
+====================================================
+
+The various python modules are installed in 
+/usr/local/lib/python3.4/dist-packages/
+
+You can delete the subfolders manually or depending on the package, you can also rely on pip or the listing of files you have recorded.
+
+sudo apt-get install python3-pip 
+sudo pip3 uninstall robofab defcon ufo2fdk ufoLib defconQt 
+
+to get rid of all the remaining files installed : (careful with rm)
+cat installed-files.txt | xargs sudo rm --verbose -vr
+
+
+
 
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 # import sys
-from distutils.core import setup
+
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -45,5 +51,13 @@
         "defconQt.representationFactories",
         "defconQt.util",
     ],
-    package_dir={"": "Lib"}
+    scripts=["trufont"],
```
anthrotype: this console script will not work on Windows (you can't expect to have the bash shell there, and python script are executable by the python launcher only if they have the `.py` extension).

IMO, it would be better to use setuptools "entry points" to create platform-specific console scripts (.exe on Windows, or executable scripts with the correct shebang on Unix):

More info at https://pythonhosted.org/setuptools/setuptools.html#automatic-script-creation

setuptools has many advantages over plain distutils. It is the recommended method to distribute Python packages in the User Guide of the [PyPA](https://packaging.python.org/en/latest/current/) (Python PAckaging Authority). 

Together with the pip installer, setuptools is now included by default in all official Python.org installers and many other distributions.

OK, enough with the advertisement.

setup.py
```diff
@@ -5,14 +5,20 @@
 try:
     import fontTools  # noqa
 except:
-    print("*** Warning: defcon requires FontTools, see:")
-    print("    fonttools.sf.net")
+    print("*** Warning: trufont requires FontTools, see:")
```
anthrotype: with setuptools you could specify, e.g.,`install_requires = ['fonttools>=3.0']` keyword in the `setup` function, so that it could grab the required packages before attempting to install trufont itself.
but you all know that ;)

n7s:
Ah yes, thanks for pointing that out. I'll suggest something more appropriate.

INSTALL.txt
```diff
@@ -1,4 +1,50 @@
...
-  python setup.py install
+Installation steps for Ubuntu/Debian
+=====================================
+
+(Note that we are using version 3.4 of python).
+
+
+sudo apt-get install -qq -y python3-pyqt5 python3-flake8
+
+git clone --depth=1 --branch=python3-ufo3 https://github.com/trufont/robofab
```
anthrotype: unless for developers, I'm not sure it's a good idea for end-users to install the trufont's dependencies from their git repositories, let alone from master/HEAD.
What if one pushes something to, say, fonttools which inadvertently breaks something else in trufont?

I think it would be better to use a `requirements.txt` file, or maybe two of them: a "stable" one with packages frozen at specific commits which are known to be working; and a 'dev-requirements.txt` file which would clone from HEAD.

This way one could do `pip install -r requirements.txt` and pip would automatically grab all the dependencies.

A cool thing about pip + requirements.txt files, is that one can specify a git URL for a package (when that is not already available on PyPI) -- even a specific commit or tag:

E.g.:
```bash
$ echo 'git+http://www.github.com/behdad/fonttools.git@894a52a90fd2670a6d01af8ef0f9967bfc682be3#egg=fonttools' > requirements.txt
$ pip install -r requirements.txt
```

This would install the fonttools package from its github repo at a specific git commit hash.

More info here: https://pip.readthedocs.org/en/1.1/requirements.html

davelab6: Using pip to install from src sounds great!

setup.py
n7s: I can't seem to find the right module attribute to import: 

entry_points={
        'console_scripts': [
            "trufont = defconQt:main"
        ]

Any ideas?

anthrotype: It's the `__main__.py` module I guess, though that one does not contain any callable `main()` function you can hook to...
Setuptools wants a function name (with no arguments) in order to make a console script wrapper. 
I don't have a computer at hand now, sorry.
I can try to help tomorrow morning. Cheers!

anthrotype: @n7s 

you need to add a `main()` function in `__main__.py` module, like this:

```python

def main():
    # indent the content of __main__.py module

if __name__ == "__main__":
    main()
```

and then you add the following `entry_points` to `setup`:

```python
entry_points={
        'console_scripts': [
            "trufont = defconQt.__main__:main"
            ],
        },
```

moyogo:
Thanks.
I modified __main__ cc479ee following @anthrotype’s advice.

@n7s Next time, it’s better if you make a new branch and use that for the pull request instead of your master branch.

setup.py
```diff
@@ -1,18 +1,26 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
adrientetar: Unneeded in Python 3.

Issue #16, opened 2015-10-15
Finish implementing fontInfo attributes


adrientetar:
I have a [stalled wip patch](https://gist.github.com/adrientetar/6f3875da84896a0f3415) for this.

pathumego:
@adrientetar can this be closed?

adrientetar:
gasp/name records not implemented to this day (though that's nothing compared to when I opened this issue).

madig:
I'd implement them but I'd first have to understand Qt's Model/View/Delegate thing.

adrientetar:
[Here's an example](https://github.com/adrientetar/defconQt/blob/master/Lib/defconQt/controls/listView.py).

PR #49, opened 2015-10-22
fontInfo: Add OS/2 tab, add class TabWidget, use load and write functions

moyogo:
- Merge patch from #16
- Refactor how tabs load and write info data
- Fix bug introduced with https://github.com/trufont/trufont/commit/63f83224f474c2c191e8cac3cb6bda8f2cb81b7d#diff-cae4a348ae503edbfb0602d0ec547d37R171
- Add unittest for most info data

### DIFF
--- a/Lib/defconQt/fontInfo.py
+++ b/Lib/defconQt/fontInfo.py
@@ -46,77 +46,167 @@ class TabDialog(QDialog):
         super(TabDialog, self).accept()
 
 
-class GeneralTab(QWidget):
-    name = "General"
+class TabWidget(QWidget):
+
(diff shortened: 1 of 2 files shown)


adrientetar:
Thanks Denis, great work. Maybe put the tests in a tests/ folder?

I want to bring OS/2 tab back into OpenType with a scrollbar but that will be done later on.

adrientetar:
Also we can have those tests be run with Travis? Can be done later on as well.

moyogo:
> Maybe put the tests in a tests/ folder?

Done.

On Travis, I'm getting some import PyQt5 error so I'll have to fix that first.

PR #51, opened 2015-10-23
Travis: run setup.py test

moyogo:


adrientetar:
Is it common to put the test runner in setup.py?

moyogo:
> Is it common to put the test runner in setup.py?

It is but it’s probably better to use `test/__init__.py`.
I'll see what I can do.

moyogo:
The TestCommand was removed and setup.py has test_suite="tests" instead. The deconQt/fontInfo_test.py was also moved to test_fontInfo.py, as it makes things easier (unittest discover doesn’t look in subfolders on its own) and test_*.py is the default pattern for discoverable unittest filename.

moyogo:
Scratch the second sentence of my last message. I was missing a __init__.py in the subfolder.

So the TestCommand was removed and setup.py has test_suite="tests" instead.

anthrotype:
:+1:

Issue #54, opened 2015-10-23
[darwin] Include NSPrincipalClass and NSHighResolutionCapable in Info.plist


davelab6:
Currently the mac package is not set up for retina displays:

<img width="1282" alt="screen shot 2015-10-23 at 14 52 48" src="https://cloud.githubusercontent.com/assets/261579/10699485/ca65d69e-7995-11e5-8c3f-b5694d40845d.png">

This can be fixed simply, by including the following 2 keys in the top of `TruFont.app/Contents/Info.plist`

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
<key>NSPrincipalClass</key>
<string>NSApplication</string>
<key>NSHighResolutionCapable</key>
<true/>
...
```

Then a hidpi mode is used:

<img width="1166" alt="screen shot 2015-10-23 at 14 51 38" src="https://cloud.githubusercontent.com/assets/261579/10699489/cfb46e9e-7995-11e5-8eb5-688609caeca9.png">

davelab6:
I'll make a PR for this shortly

PR #58, opened 2015-10-23
Add Lib/TruFont.spec

davelab6:
Update Lib/build.sh to use spec file, to fix https://github.com/trufont/trufont/issues/54


Lib/TruFont.spec
```diff
@@ -0,0 +1,37 @@
...
+          a.datas,
+          name='TruFont',
+          debug=False,
+          strip=None,
+          upx=True,
+          console=False, 
+		  icon='defconQt/resources/app.ico')
+app = BUNDLE(exe,
+             name='TruFont.app',
+             icon='defconQt/resources/app.ico',
```
adrientetar: On OSX you need to use the icns icon, not ico.

Lib/build.sh
```diff
@@ -1 +1,7 @@
-pyinstaller --noconsole --name "TruFont" --onefile -i "defconQt/resources/app.ico"  defconQt/__main__.py
+pyinstaller \
+	--noconsole \
+	--name "TruFont" \
+	--onefile \
+	--icon="defconQt/resources/app.ico" \
+	--osx-bundle-identifier=io.github.trufont \
+	TruFont.spec;
```
adrientetar: Does this work with the other platforms? Maybe this file should be changed into a python script.

Lib/TruFont.spec
davelab6: Thanks!

Lib/build.sh
davelab6: It won't work on windows; I've reverted this.

PR #86, opened 2015-10-27
[fontView] s/Mark/Flag/g in UI strings only

### LINKED ISSUE #80: Another name for "Mark Color" ?
A user complained that "Mark Color" made he think it had something to do with diacritical marks. Perhaps we could come up with a better, less ambiguous name for this option.

davelab6:
Resolves #80


adrientetar:
I'd rather we get the renames done in UFO4.

PR #123, opened 2015-11-07
meta: add custom glyphList support

adrientetar:
@moyogo What do you think?

This adds support of custom glyph lists in the form `A;0041` as in AGL (not AGLFN which is backwards and has an additional field). 
I was thinking, since glyph names in UFO are unboun, you could very much have ; in a glyph name.

### DIFF
--- a/Lib/defconQt/__main__.py
+++ b/Lib/defconQt/__main__.py
@@ -1,9 +1,10 @@
-from defconQt.objects.defcon import TFont
 from defconQt import representationFactories
 from defconQt import icons_db  # noqa
 from defconQt.fontView import Application, MainWindow
+from defconQt.objects.defcon import TFont
 import sys
 import os
+from PyQt5.QtCore import QSettings
 from PyQt5.QtGui import QIcon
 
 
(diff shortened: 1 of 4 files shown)


davelab6:
> I was thinking, since glyph names in UFO are unboun, you could very much have ; in a glyph name.

To support `A;swash;0041` then make the final `;` the separator?

adrientetar:
Okay let's mimic AGL requirements for now, I don't esp. like the idea of parsing arbitrary strings.

moyogo:
https://github.com/trufont/trufont/commit/eb6d3fe775e793f7007703ed7489e87827ba0a47#diff-d91809653932fd3783feb1ab7565d12dR3 could share the same Unicode value regex part as https://github.com/trufont/trufont/blob/master/Lib/defconQt/fontView.py#L141
We could restrict it to valid values less or equal to 10FFFF and exclude the surrogate pairs D800...DFFF.
@adrientetar should we add this to util/glyphlist.py and import it in fontView.py? Or should this be in a new file like util/uniData.py?

http://unifiedfontobject.org/versions/ufo3/glif.html#glyph says:
> The name of the glyph. This must be at least one character long. Different font specifications, such as OpenType, often have their own glyph name restrictions. Authoring tools should not make assumptions about the validity of a glyph's name for a particular font specification.

One could use invalid PS characters, like the semicolon as Dave suggests, or even non ASCII characters in glyph names. It gets tricky when saving as UFO2, it would need a way to translate that with what UFO2 applications accept as glyph names.

adrientetar:
@moyogo That can go in glyphList.

PR #134, opened 2015-11-09
Avoid crashing the application in case there's not yet any config file

felipesanches:
Avoid crashing the application in case there's not yet any config file, by passing a default value to the glyphList entry in the TruFont settings.


adrientetar:
I thought the default default value was an empty string already. I'll check.

PR #140, opened 2015-11-13
Add requirements.txt

teddywing:
Having to install dependencies manually can be troublesome. Not only does it take longer, but updating packages and version locking becomes difficult.

This change adds a `requirements.txt` so that all Python packages (except for PyQt5) can be installed via `pip` with a single command.

Update the `INSTALL.md` file with new instructions that use the `requirements.txt` file. Also add a note to this file about how to use Virtualenv with the project (because of the way PyQt5 needs to be installed).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+fonttools==3.0
+git+https://github.com/trufont/defcon.git@python3-ufo3#egg=defcon
+git+https://github.com/trufont/robofab.git@python3-ufo3#egg=robofab
+git+https://github.com/trufont/ufo2fdk.git@python3-ufo3#egg=ufo2fdk
+PyInstaller==3.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,9 @@ dist: trusty
 language: python
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq -y python3-pyqt5 python3-flake8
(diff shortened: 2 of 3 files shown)


davelab6:
Awesome work @teddywing :)

adrientetar:
Thanks!

moyogo:
Cool!

requirements.txt
```diff
@@ -0,0 +1,5 @@
+fonttools==3.0
```
anthrotype: The version on PyPI is old and is not the latest HEAD. I would recommend to pip install fonttools from the git repo instead

moyogo:
@anthrotype @davelab6 time for a new fonttools release?

requirements.txt
davelab6: Its the latest release, https://github.com/behdad/fonttools/releases

anthrotype: I know Dave, but it's 3 months old and quite behind the current HEAD. OK,
we shall discuss on the upstream repo whether we need to tag a new 3.1
release.
Il 13/nov/2015 08:13, "Dave Crossland" <notifications@github.com> ha
scritto:

adrientetar:
We need to use HEAD, we depend on changes that were made recently. Does someone want to patch this?

teddywing:
Sure, I can make the update

teddywing:
Patch created as #141

PR #141, opened 2015-11-13
requirements.txt: Use current Git HEAD version of fonttools

teddywing:
The 3.0 release previously specified was from August 31 and a number of
changes have been made since then. TruFont depends on these new changes,
but since 3.0 is the most recent release of FontTools, we need to pull
the changes from its Git repository. Pinning it to the latest commit as
of now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-fonttools==3.0
-git+https://github.com/trufont/defcon.git@python3-ufo3#egg=defcon
-git+https://github.com/trufont/robofab.git@python3-ufo3#egg=robofab
-git+https://github.com/trufont/ufo2fdk.git@python3-ufo3#egg=ufo2fdk
+git+https://github.com/behdad/fonttools.git#egg=fonttools
+git+https://github.com/trufont/defcon.git@559c8f6bde1629d13f4e8a69e1f6f2494d977c32#egg=defcon
+git+https://github.com/trufont/robofab.git@f646d1f36b7294ffd70e5b4d4e1c7c7155852585#egg=robofab
+git+https://github.com/trufont/ufo2fdk.git@0a57f2474ff65d5877b28bd063a249bee1a18ec6#egg=ufo2fdk
 PyInstaller==3.0


requirements.txt
```diff
@@ -1,4 +1,4 @@
-fonttools==3.0
+git+https://github.com/behdad/fonttools.git@7f5d5454fc1332815a42ba8b9fb666981680bce4#egg=fonttools
```
anthrotype: good idea to freeze to a specific commit in time, instead of pulling it from HEAD.
I'd do it for all other requirements as well, so that one knows at any point in time the stable, supported state of all the dependencies.

teddywing: Good idea. Does it make sense to make that change in this pull request?

anthrotype: I guess so

davelab6: Yes, and rebase to squash the commits into 1 :)

anthrotype: of course, it'd be even better if the commits were tagged, instead of being referenced by their hash. But sometimes we don't have control over this.

adrientetar:
Please use head from fontTools. Even if something was to break, we can signal it to the maintainers immediately and they respond fast usually.

davelab6:
I agree with adrien, HEAD is better for development

anthrotype:
@adrientetar that's true. however, having a `requirements.txt` file with precise commit hashes would allow to replicate the same environment by different people on different machines. That might help to debug issues, for example. Anyway, feel free to do as you think it's better.

anthrotype:
or you could have two `dev-requirements.txt` and `stable-requirements.txt` files maybe?

anthrotype:
never mind, I don't want to block this. Let's go for the HEAD for now, we can think about having a separate requirements.txt file which would contain precise commit hashes (and maybe linking this to a given tagged trufont release).

teddywing:
@adrientetar I'm happy to change to use the fontTools HEAD, but I agree with @anthrotype that using a specific commit hash is helpful for keeping all environments in sync. Will make the change as requested but might be worth revisiting.

teddywing:
Updated to use fontTools@HEAD and commit hashes for the other Git dependencies

adrientetar:
Merged, thanks.

PR #144, opened 2015-11-19
Use QPlainTextEdit for copyright and license in font info

khaledhosny:
These two fields typically have longer text that doesn’t into a single
line and using QLineEdit makes editing them cumbersome.

### DIFF
--- a/Lib/defconQt/fontInfo.py
+++ b/Lib/defconQt/fontInfo.py
@@ -68,6 +68,20 @@ class TabWidget(QWidget):
         else:
             setattr(font.info, dst, None)
 
+    def loadMultilineString(self, font, src, dst):
+        value = getattr(font.info, src)
+        if value is not None:
+            setattr(self, dst + "Edit", QPlainTextEdit(value, self))
+        else:
+            setattr(self, dst + "Edit", QPlainTextEdit(None, self))
+
+    def writeMultilineString(self, font, src, dst):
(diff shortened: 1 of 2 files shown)


khaledhosny:
It is always nice to have unit tests that tell you when you mess things up, really appreciated :)

adrientetar:
Cool, can you change loadMultiline/writeMultiline/checkMultiline to loadMultilineString & al. Also can you use loadMultilineString for font.info.note as well?

Also I know that it was multiline strings for copyright & friends in FontForge but is it common in end-user apps to show copyright/trademark strings as multiline?

khaledhosny:
> Cool, can you change loadMultiline/writeMultiline/checkMultiline to loadMultilineString & al.

I used that originally, but it felt too long and multiline is clearly a
string, no strong opinion either way.

> Also can you use loadMultilineString for font.info.note as well?

I didn’t do that as I was using setMaximumHeight and did not want to
change the note widget, but I turns out I don’t need it (no idea why I
felt the need!), so done now.

> Also I know that it was multiline strings for copyright & friends in
> FontForge but is it common in end-user apps to show
> copyright/trademark strings as multiline?

GNOME font viewer shows the copyright as multiline string, no license is
shown and no other apps here that expose this.

adrientetar:
Thanks.

moyogo:
:+1: Thanks!

PR #148, opened 2015-11-30
Show old anchor name in the rename dialog

khaledhosny:
Helps when one wants to prefix or suffix the name or have a so short
memory span like me and immediately forgets what the old name was. The
text is highlighted, so shouldn’t prevent one from typing a new name
right away as before.

### DIFF
--- a/Lib/defconQt/glyphView.py
+++ b/Lib/defconQt/glyphView.py
@@ -115,10 +115,10 @@ class GotoDialog(QDialog):
 
 class AddAnchorDialog(QDialog):
 
-    def __init__(self, pos=None, parent=None):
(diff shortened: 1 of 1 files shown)


adrientetar:
Thanks. I'm touching this just now, give me a few days.

davelab6:
@adrientetar ping

adrientetar:
This has rotten because it was submitted when I was working on the new glyphView. I'll merge it manually.

adrientetar:
Btw this PR makes the incorrect assumption that name has a len ≡ anchor is preexistant. This isn't true, an anchor may be unnamed or have a zero-length name.

adrientetar:
599cca2293a0f343719f5bb3c772148ecdefafc4

Issue #115, opened 2015-11-05
Keep width in sync across layers?


adrientetar:
cc @graphicore

adrientetar:
@twardoch Does FLVI do anything to keep glyph width in sync across layers? e.g. when you space font, does it change glyph width for all layers?

adrientetar:
Alternatively we could always paint the horizontal metrics of the foreground glyph.

adrientetar:
As of 8c063ed we always use the default layer for metrics.

Issue #14, opened 2015-10-15
Write a tracking undo manager


adrientetar:
That makes use of defcon `undoManager` calls as well. Then cleanup current hackish system in glyphView.
- should have undo/redo notifications for observers (GC)

adrientetar:
@graphicore, feel free to work on this if you want to.

graphicore:
Yeah, there's no defcon undoManager anymore in defcon ufo3. So this is a major task that requires heavy work on defcon. At least if you want to have the real deal, in terms of a command based undo/redo queue. I'd also prefer to do this in sync with @typesupply. And it will be time consuming.

adrientetar:
I noticed glyphs has glyph.beginUndo() and glyph.endUndo(), this seems sensible API.

adrientetar:
Also currently we stack an undo state on each mouse click even if no modification ends up being done which creates duplicates &ndash; something we ought to fix too.

typesupply:
> I noticed glyphs has glyph.beginUndo() and glyph.endUndo(), this seems sensible API.

I started an undo manager a long time ago in Defcon but ran into some issues. Undo/redo is usually presented to the user at the application level and the application level usually has its own undo manager (ie NSUndoManager in Cocoa). The issue became: where does the state stack live? App level? Object level? Both? If both, how do they stay in sync and WHY BOTH!? It made my head hurt so I gave up.

I'm still not sure. I guess it might be useful for defcon to have an optional undo manager that would exist for each object. This would let purely Pythonic environments, such as direct object scripting, to have undo support. Then again, is undo really necessary in that kind of environment?

Anyway, off the top of my head, an object would need to be able to do the following:

- store an initial state
- store a state with a description of the action performed before the state was saved
- restore from a previous state

The serialization support that we discussed would do most of this work. The big issue would be the stack manager.

adrientetar:
> Then again, is undo really necessary in that kind of environment?

It can be useful if you are doing scripting embedded into an app. Push an undo for your script that smoothes up Béziers or something, and if you don't like the result you can revert.

I would say the undoStack should live at application level but we should leave the possibility to attach whatever stack no matter whether if it's application-level or python.

Then you could pass arguments e.g.:

```
font.registerUndoStack(undoStack, undoMethod="undo", redoMethod="redo")
```

> This would let purely Pythonic environments, such as direct object scripting, to have undo support.

[...]

PR #150, opened 2015-12-03
glyphView_ng

### LINKED ISSUE #11: Measuring tool does not go away
I really like the triangle visual of the measuring tool, but it doesn't go away when I switch to a different tool: the triangle from the last thing I measured remains in the glyph window. In order to get it to go away, I have to switch to a different glyph, and then switch back.

### LINKED ISSUE #18: Refactor tools into proper & modular objects
I have a [stalled wip patch](https://gist.github.com/adrientetar/709e303c01cab61d0e0c) for this. Also make the toolbar a glyphView elements, similar to how the layers dropdown is handled.

### LINKED ISSUE #52: Jumping to glyph resets tool, but not active toolbar button
It goes back to selection tool but doesn't change toolbar button.

I tend to think we shouldn't really change the active tool in the first place.

cc @graphicore

### LINKED ISSUE #69: Include booleanOperations
To include https://github.com/typemytype/booleanOperations there are a couple of things that seem obvious to me:

- [ ] Update the module to work on python 2 and 3 (@anthrotype is already working on this :)
- [ ] set up PyInstaller to correctly package the C++ part 
- [ ] a convenience API (similar to http://doc.robofont.com/api/mojo/mojo-tools/)
- [ ] buttons on the inspector (similar to https://github.com/typemytype/pathOperatorRoboFontExtension)

### LINKED ISSUE #128: Select Only Points with Alt+Drag Select
In GlyphView if I drag select area, it selects both points and handles. I suggest `alt+Drag Select` to select only points, without handles.

### LINKED ISSUE #135: Harmonize usage of SizeGripItem widget
Adrien had requested a Qt widget that provides a set of resizing handles. One specific aspect of the request was that it should be generic so we could use it for both resizing pixmaps and for resizing glyph curves. I did it by porting the C++ reference implementation that Adrien pointed out here: https://github.com/felipesanches/sizegripitem_python

Over the weekend, though, Adrien merged a portion of that code into TruFont, but including some changes to use it for scaling pixmaps. The issue is that these changes break the overall structure and generality of the widget. I've spent quite a large ammount of time today trying to re-harmonize the two usages of the widget. But instead of simply doing it quietly, I preffer to explain here the design of the widget so that we can see if we agree on these terms.

The glyph is built with 2 classes: **SizeGripItem** which is the main one and which instantiates several **ResizeHandleItem**s (which are the control nodes of the resizer widget).

The SizeGripItem class holds a reference to a resizer object that must be passed to it in its constructor.

The purpose of this resizer object is to make the widget generic. The resizer object is responsible for providing a resize() method that does whatever specific-purpose operation that a specific tool may need. So, in the case of the pixmap usage, this would be performing the resizing operation on the selected pixmap. In the glyph outline usage, it means re-evaluating the coordinates of all curve nodes that belong to the selection.

Adrien's implementation with pixmaps simply does not use this mechanism and the changes affect the usage of the SizeGripItem for glyph curves resizing (which is my current work in progress). I'd like to discuss it here so that we reach a consensus on the correct way of handling this in a generic way.

adrientetar:
Kind of a lot of stuff here, and some things aren't finished yet. What this brings up:

- a new glyphView with manual on-widget painting, scaling and coordinates correspondance
- modular tools and buttons
- correct and fast algorithms to move/delete any selection
- correct and fast algorithm to filter and extract glyph selection, selecting points draws selected path
- constant on-canvas inverse scaling
- glyph-specific undo stack with unchanged state pruning
- remove overlap (script + GUI)
- improved selected state support, impl. for Anchor, Component too &ndash; added Object.SelectionChanged notification (glyph watches for Anchor/Component notifications)
- new knife tool icon
- split out parts to other files
- off curve points have their selected state set, a new alt. selection mode (w Alt) does not select them
- possibility to cycle between glyphs with End or Home
- possibility to cycle between contour points with \> or \<
- more

TODO:

- [x] correctly subscribe to and handle layerSet changes in the window controller
- [x] create anchors/components
- [x] finish impl. menus w save, goto&al, display, zoom, toolbar lock&hellip;
- [x] correct undoManager indexings
- [x] split things to a dict in _itemAt
- [x] finish Pen tool and correct closed contour sliding
- [x] sort out focus policies and propagation
- [x] remove glyphView and the older OutlineInformation_ pen, remove globs, cleanup

Later:

- [ ] tools access in scripting w/o importing
- [ ] implement layer/display settings + layer color in the inspector tool window (w tree widget)
- [ ] store per-layer, window independent display settings
- [ ] make point sizes and drawing parameters tunable

Glyph view should be in pretty good shape once that is done. After that only inspectorWindow and fontInfo will remain to be touched up still.

Closes #11.
Closes #18.
Closes #52.
Closes #69.
Closes #128.
Closes #135 (obsoleted).
#14 is improved on although we may want a QUndo-based system later on.
#115 is resolved although the synchronisation of width across layers remains a question.


davelab6:
Awesome!!! Well done Adrien!! :)

I'm happy you slipped remove overlap in there too :)

Lib/defconQt/objects/glyphDialogs.py
```diff
@@ -0,0 +1,239 @@
+from PyQt5.QtCore import Qt
+from PyQt5.QtGui import QEvent
```
moyogo: This should be
```python
from PyQt5.QtCore import Qt, QEvent
```

pathumego:
:+1:  Thanks @adrientetar !!!

PR #206, opened 2016-01-23
Upgrade to LGPLv3

### LINKED ISSUE #4: License
What is the license of Trufont?

davelab6:
Closes #4


adrientetar:
5093bf7f1f54e684d3c81ebffae548b10a7df4a5

davelab6:
Why didn't you take the clarification of the COPYRIGHT file?

adrientetar:
It's more words and more confusing but without providing more actual information.

- You don't need a commercial license of Qt to be able to use the LGPL license so why would you talk about that
- I don't want to get into the details of saying "you need to purchase additional license", I just put a hint here and the one interested can go look at PyQt license on the official website (the one who's concerned about license will end up looking at it anyway)
- You are too fussy with your wording, e.g. `You may purchase additional licenses` ... ` to get additional permission so that you may distribute` ... `but without those additional paid licenses you must distribute` ... `under the terms of the GNU GPL v3, to comply with the PyQt5 GPLv3`  
Why are you superfluously repeating the same words in the same sentence? Reading COPYRIGHT is not a pastime

davelab6:
> You don't need a commercial license of Qt to be able to use the LGPL license so why would you talk about that

You need a commercial license of Qt to ship a totally proprietary app; the LGPL carries a lot of obligations, such as allowing end users to upgrade their Qt. 

> I don't want to get into the details of saying "you need to purchase additional license", I just put a hint here and the one interested can go look at PyQt license on the official website (the one who's concerned about license will end up looking at it anyway)

I think a hint is confusing and its better to be explicit. 

> You are too fussy with your wording .. Why are you superfluously repeating the same words in the same sentence? Reading COPYRIGHT is not a pastime

Fair :) How about this?

https://github.com/trufont/trufont/compare/master...davelab6:copyright-clarification?expand=1

PR #217, opened 2016-01-28
Integrate py2app

### LINKED ISSUE #90: How to build packages?
https://github.com/pyinstaller/pyinstaller/issues/1309 suggest py2app could be used for the OS X packaging instead, so that drag'n'drop to dock, Open With, and so on will work

screen shot 2015-10-29 at 06 42 56

teddywing:
## Requesting assistance

1. The glyph view button borders get messed up when building with py2app and I can't figure out why. Can someone take a look and help me out?

    screen shot 2016-01-25 at 5 26 54 pm

2. The resulting application bundle is twice as large as the one built by PyInstaller. I'm a little stumped here too.

Purpose:
--------

This change switches the packaging tool on Mac from PyInstaller to
Py2app in order to take advantage of Py2app's support for ARGV
emulation. This allows us to open files by dragging them onto the
application.

With Adrien's change ac63e80c57b37527f331b209f88d583deeeabcd3, dragging
files onto the application will open files even if the application is
already launched.

To Build:
---------

Run

    $ sh Lib/build.sh

This is the same as:

    $ python setup.py py2app

Only use py2app if we run the setup file on a Mac. We also have some
pre- and post-build instructions that need to be run, and these are only
executed when on a Mac and the `py2app` argument is passed to `setup.py`
(so they don't get mistakenly run on `python setup.py install`).

If the build script is run on a non-Mac platform, PyInstaller will be
used to build the application bundle as before.

Additional Information
----------------------

* setup.py: Uncommented the `sys` import as I needed it to determine the
  current platform and whether we're running `py2app`.
* Globs are used in the Qt framework paths to allow developers/packagers
  to use different versions of Qt5. I did this because I'm running OS X
  10.8 and couldn't get Qt 5.5 to build on my machine, so I'm using
  homebrew/versions/qt52. Homebrew is assumed.
* The `libqsvgicon.dylib` dylib must be added manually or our glyph view
  toolbar icons don't get rendered.
* The `libqcocoa.dylib` library must be added or we get this error when
  trying to start the application:

      This application failed to start because it could not find or load the Qt platform plugin "cocoa".

      Reinstalling the application may fix this problem.
      Abort trap: 6

* We supply a `qt.conf` file to specify the location of the Qt plugins
  directory. This allows us to specify where to find the dylibs we
  manually copied into the application bundle.

Info.plist
----------

The Info.plist file generated for the application bundle has been
extended and modified.

* `NSHighResolutionCapable=YES` was kept
* `LSBackgroundOnly=NO` was removed because not setting it turns it off
  by default
* Include the `CFBundleIdentifier` copied from `TruFont.spec`
* Set `NSHumanReadableCopyright` to empty string. Otherwise by default
  Py2app will insert this key for you whether you like it or not with
  the value "Copyright not specified".
* Add the `CFBundleDocumentTypes` key to declare TruFont as an app
  capable of opening *.ufo files. This allows us to select TruFont as
  the default app to open one or all files of this type using the
  "Open with…" functionality throughout OS X.

Known Issues
------------

* The button borders in the glyph view toolbar are wrong. They're square
  when they should be invisible with gradiented backgrounds. Not sure
  how to correct this.
* This application bundle is about 20 Mb larger than the previous one
  generated by PyInstaller. Not sure what's going on there.

Fixes #90.

### DIFF
--- a/Lib/build.sh
+++ b/Lib/build.sh
@@ -1,1 +1,5 @@
-pyinstaller --onefile TruFont.spec
+if [ $(uname) == 'Darwin' ]; then
+	python setup.py py2app
+else
+	pyinstaller --onefile TruFont.spec
+fi
--- a/qt.conf
+++ b/qt.conf
@@ -1,0 +1,2 @@
+[Paths]
+Plugins = PlugIns
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,11 @@
 #!/usr/bin/env python
-# import sys
+
+from glob import glob
+import os
 from setuptools import setup
+import shutil
+from subprocess import call
+import sys
 
 try:
     import fontTools  # noqa
@@ -21,8 +26,48 @@ except:
           "see:")
     print("    https://github.com/trufont/defcon")
 
+
+name = 'TruFont'
+app = []
+options = {}
+data_files = []
+setup_requires = []
+build_for_mac = sys.platform == 'darwin' and sys.argv[1] == 'py2app'
+
+if build_for_mac:
+    options['py2app'] = {
+        'argv_emulation': True,
+        'iconfile': 'Lib/defconQt/resources/app.icns',
+        'includes': ['sip', 'PyQt5', 'defconQt'],
+        'frameworks': [
+            glob('/usr/local/Cellar/qt5*/5.*/plugins/platforms/libqcocoa.dylib')[0],
+            glob('/usr/local/Cellar/qt5*/5.*/plugins/iconengines/libqsvgicon.dylib')[0],
+        ],
+        'qt_plugins': ['*'],
+        'plist': {
+            'CFBundleIdentifier': 'io.github.trufont',
+            'CFBundleDocumentTypes': [
+                {
+                    'CFBundleTypeExtensions': [
+                        'ufo',
+                    ],
+                    'CFBundleTypeName': 'Unified Font Object',
+                    'CFBundleTypeRole': 'Editor',
+                    'LSTypeIsPackage': True,
+                }
+            ],
+            'NSHighResolutionCapable': 'True',
+            'NSHumanReadableCopyright': '',
+        }
+    }
+
+    app.append('Lib/defconQt/__main__.py')
+    data_files.append('qt.conf')
+    setup_requires.append('py2app')
+
+
 setup(
-    name="defconQt",
+    name=name,
     version="0.3.0",
     description="TruFont, a cross-platform font editor. Includes a set of Qt "
                 "objects for working with font data.",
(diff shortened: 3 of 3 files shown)


adrientetar:
Thanks! I'll have a look when I have OSX available.

requirements.txt
```diff
@@ -6,3 +6,4 @@ setuptools>=18.0
 -e git+https://github.com/typesupply/extractor.git@ufo3#egg=extractor
 -e git+https://github.com/typesupply/featools.git#egg=featools
 -e git+https://github.com/googlei18n/ufo2ft.git#egg=ufo2ft
+py2app==0.9
```
adrientetar: I'd rather we don't put this in the requirements because much like Pyinstaller it's not required to run the program. Both are well-packaged and easy to install.

teddywing:
Looks like I have a Travis error. Do you think I should remove py2app from `requirements.txt`? Would that fix it?

requirements.txt
teddywing: Ok I can remove it

adrientetar:
There's an error ` error: option --single-version-externally-managed not recognized` on py2app install, so yeah I guess removing it would prevent the failure.

teddywing:
Removed the `requirements.txt` change

adrientetar:
> 1. The glyph view button borders get messed up when building with py2app and I can't figure out why. Can someone take a look and help me out?
2. The resulting application bundle is twice as large as the one built by PyInstaller. I'm a little stumped here too.

I'll take a look at these soon what I have OSX available.

teddywing:
Thanks!

adrientetar:
Sorry for delay, I intend to tackle this in a few days and hopefully get a new version out.

teddywing:
No problem, thanks for the update

adrientetar:
I want to make a new release and I've spent the afternoon working on getting Qt build to work (which I think would fix the shortcomings of this PR and hopefully could work for all platforms easily...), but still have failures so far.

adrientetar:
@teddywing Okay I pushed a new OSX build [here](https://github.com/trufont/trufont/releases/tag/0.3.0), it still seems rather large package (~40MB) but it has touchups on the platform integration side. WDYT?

adrientetar:
Okay I've had good feedback with the Qt build system. Keeping this for reference, thanks.

teddywing:
Sorry, I know it's been a while, but I just wanted to thank you for reviewing this Adrien

adrientetar:
Haha, thanks for the PR!

davelab6:
> I've had good feedback with the Qt build system

Is there a URL for that?

With 0.4.0, dragging a UFO to the TruFont dock item doesn't work, dragging and dropping a UFO into a empty new Font window doesn't work.... Alas!

PR #221, opened 2016-02-01
Metrics window: make wrapping a simple checkable action.

madig:
https://developer.gnome.org/hig/stable/menus.html.de: "Two linked
actions can be combined into a single menu item, by changing the label
when the item is selected. For example, a Play item may change to Pause.
However, only use this type of item when actions are logical opposites
which are obvious to users. Likewise, do not use this technique for
settings - use check boxes or radio buttons instead."

PS: did you rewrite the last commit? PLEASE don't do that for a master branch that others pull from!


adrientetar:
You use two toggles in your code, only one is needed for one setting. The toggle is the QAction, see a5dc5b40153171a49506e802862f489dc5f59ea7.

adrientetar:
> PS: did you rewrite the last commit? PLEASE don't do that for a master branch that others pull from!

Sorry about that.

PR #224, opened 2016-02-01
Translation support

madig:
I'd like to get this in, but maybe without my example translation. As long as the HIG/UI has not stabilized, it is only useful for testing.

Known problems:
- Translations don't load when starting via `trufont` -- package into resource file?
- `pylupdate5` gets confused when strings contain non-ASCII characters. Either avoid or escape them (quotation marks!) or attach them afterwards (ellipsis).

Notes:
- When sizing elements like buttons and columns, don't set a fixed width based on the length of English words. I'd fix fontView.py's glyphListButton if I knew why sizeHint() makes it so wide.

### DIFF
--- a/Lib/defconQt/__main__.py
+++ b/Lib/defconQt/__main__.py
@@ -1,8 +1,11 @@
 from defconQt import __version__, representationFactories
 from defconQt import icons_db  # noqa
 from defconQt.fontView import Application
-from PyQt5.QtCore import QCommandLineParser, QSettings
+from PyQt5.QtCore import (
+    QCommandLineParser, QSettings, QTranslator, QLocale,
+    QLibraryInfo)
 from PyQt5.QtGui import QIcon
+from PyQt5.QtWidgets import QApplication
 import os
 import sys
 
@@ -17,13 +20,29 @@ def main():
     app.setApplicationName("TruFont")
     app.setApplicationVersion(__version__)
     app.setWindowIcon(QIcon(":/resources/app.png"))
+
+    # Qt's translation for itself. May not be installed.
+    qtTranslator = QTranslator()
+    qtTranslator.load("qt_" + QLocale.system().name(),
+                      QLibraryInfo.location(QLibraryInfo.TranslationsPath))
+    app.installTranslator(qtTranslator)
+
+    appTranslator = QTranslator()
+    appTranslator.load("trufont_" + QLocale.system().name(),
+                       os.path.dirname(os.path.realpath(__file__)) +
+                       "/resources")
+    app.installTranslator(appTranslator)
+
     app.loadGlyphList()
     # parse options and open fonts
     parser = QCommandLineParser()
-    parser.setApplicationDescription("The TruFont font editor.")
+    parser.setApplicationDescription(QApplication.translate(
+        "Command-line parser", "The TruFont font editor."))
     parser.addHelpOption()
     parser.addVersionOption()
-    parser.addPositionalArgument("files", "The UFO files to open.")
+    parser.addPositionalArgument(QApplication.translate(
+        "Command-line parser", "files"), QApplication.translate(
+        "Command-line parser", "The UFO files to open."))
     parser.process(app)
     args = parser.positionalArguments()
     if not len(args):
--- a/Lib/defconQt/featureTextEditor.py
+++ b/Lib/defconQt/featureTextEditor.py
@@ -17,27 +17,24 @@ class MainEditWindow(QMainWindow):
         self.editor = FeatureTextEditor(self.font.features.text, self)
         self.resize(600, 500)
 
-        fileMenu = QMenu("&File", self)
-        fileMenu.addAction("&Save…", self.save, QKeySequence.Save)
+        fileMenu = QMenu(self.tr("&File"), self)
+        fileMenu.addAction(self.tr("&Save…"), self.save, QKeySequence.Save)
         fileMenu.addSeparator()
-        fileMenu.addAction("&Reload From Disk", self.reload)
-        fileMenu.addAction("E&xit", self.close, QKeySequence.Quit)
+        fileMenu.addAction(self.tr("&Reload From Disk"), self.reload)
+        fileMenu.addAction(self.tr("&Close"), self.close, QKeySequence.Quit)
         self.menuBar().addMenu(fileMenu)
 
         self.setCentralWidget(self.editor)
-        self.setWindowTitle(font=self.font)
+        self.setWindowTitle(self.tr("Font Features - {0} {1}").format(
+            self.font.info.familyName, self.font.info.styleName))
         self.editor.modificationChanged.connect(self.setWindowModified)
 
     def setWindowTitle(self, title="Font Features", font=None):
-        if font is not None:
-            puts = "[*]%s – %s %s" % (
-                title, self.font.info.familyName, self.font.info.styleName)
-        else:
-            puts = "[*]%s" % title
-        super(MainEditWindow, self).setWindowTitle(puts)
+        super(MainEditWindow, self).setWindowTitle("[*]" + title)
 
     def _fontInfoChanged(self, notification):
-        self.setWindowTitle(font=self.font)
+        self.setWindowTitle(self.tr("Font Features - {0} {1}").format(
+            self.font.info.familyName, self.font.info.styleName))
 
     def closeEvent(self, event):
         if self.editor.document().isModified():
@@ -45,12 +42,12 @@ class MainEditWindow(QMainWindow):
             closeDialog = QMessageBox(
                 QMessageBox.Question,
                 name,
-                "Save your changes?",
+                self.tr("Save your changes?"),
                 QMessageBox.Save | QMessageBox.Discard | QMessageBox.Cancel,
                 self
             )
             closeDialog.setInformativeText(
-                "Your changes will be lost if you don’t save them."
+                self.tr("Your changes will be lost if you don’t save them.")
             )
             closeDialog.setModal(True)
             ret = closeDialog.exec_()
--- a/Lib/defconQt/fontInfo.py
+++ b/Lib/defconQt/fontInfo.py
@@ -38,7 +38,7 @@ class TabDialog(QDialog):
         mainLayout.addWidget(buttonBox)
         self.setLayout(mainLayout)
 
-        self.setWindowTitle("Font Info – %s %s" % (
+        self.setWindowTitle(self.tr("Font Info – {0} {1}").format(
             self.font.info.familyName, self.font.info.styleName))
 
     def accept(self):
@@ -179,23 +179,25 @@ class TabWidget(QWidget):
 class GeneralTab(TabWidget):
 
     def __init__(self, font, parent=None):
-        super().__init__(parent, name="General")
+        super().__init__(parent)
+        self.name = self.tr("General")
+
         mainLayout = QGridLayout(self)
 
-        familyNameLabel = QLabel("Family name:", self)
-        styleNameLabel = QLabel("Style name:", self)
-        designerLabel = QLabel("Designer:", self)
-        designerURLLabel = QLabel("Designer URL:", self)
-        manufacturerLabel = QLabel("Manufacturer:", self)
-        manufacturerURLLabel = QLabel("Manufacturer URL:", self)
-        copyrightLabel = QLabel("Copyright:", self)
-        licenseLabel = QLabel("License:", self)
-        licenseURLLabel = QLabel("License URL:", self)
-        trademarkLabel = QLabel("Trademark:", self)
+        familyNameLabel = QLabel(self.tr("Family name:"), self)
+        styleNameLabel = QLabel(self.tr("Style name:"), self)
+        designerLabel = QLabel(self.tr("Designer:"), self)
+        designerURLLabel = QLabel(self.tr("Designer URL:"), self)
+        manufacturerLabel = QLabel(self.tr("Manufacturer:"), self)
+        manufacturerURLLabel = QLabel(self.tr("Manufacturer URL:"), self)
+        copyrightLabel = QLabel(self.tr("Copyright:"), self)
+        licenseLabel = QLabel(self.tr("License:"), self)
+        licenseURLLabel = QLabel(self.tr("License URL:"), self)
+        trademarkLabel = QLabel(self.tr("Trademark:"), self)
         # TODO: give visual feedback of input data validity using QLineEdit
         # lose focus event
         # http://snorf.net/blog/2014/08/09/using-qvalidator-in-pyqt4-to-validate-user-input/ # noqa
-        versionLabel = QLabel("Version:", self)
+        versionLabel = QLabel(self.tr("Version:"), self)
         versionDotLabel = QLabel(".", self)
         self.loadString(font, "familyName", "familyName")
         self.loadString(font, "styleName", "styleName")
@@ -230,7 +232,7 @@ class GeneralTab(TabWidget):
         mainLayout.addWidget(versionDotLabel, 6, 2)
         mainLayout.addWidget(self.versionMinorEdit, 6, 3)
 
-        dateCreatedLabel = QLabel("Date created:", self)
+        dateCreatedLabel = QLabel(self.tr("Date created:"), self)
         dateTime = QDateTime()
         # dateTime.fromString(font.info.openTypeHeadCreated, "yyyy/MM/dd
         # hh:mm:ss") # XXX: why does this not work?
@@ -284,15 +286,18 @@ class GeneralTab(TabWidget):
 class MetricsTab(TabWidget):
 
     def __init__(self, font, parent=None):
-        super().__init__(parent, name="Metrics")
+        super().__init__(parent)
+        self.name = self.tr("Metrics")
+
         mainLayout = QGridLayout()
 
-        styleMapFamilyLabel = QLabel("Style map family name:", self)
+        styleMapFamilyLabel = QLabel(self.tr("Style map family name:"), self)
         self.styleMapFamilyEdit = QLineEdit(font.info.styleMapFamilyName, self)
 
-        styleMapStyleLabel = QLabel("Style map style name:", self)
+        styleMapStyleLabel = QLabel(self.tr("Style map style name:"), self)
         self.styleMapStyleDrop = QComboBox(self)
-        items = ["None", "Regular", "Italic", "Bold", "Bold Italic"]
+        items = [self.tr("None"), self.tr("Regular"), self.tr("Italic"),
+                 self.tr("Bold"), self.tr("Bold Italic")]
         self.styleMapStyleDrop.insertItems(0, items)
         sn = font.info.styleMapStyleName
         if sn == "regular":
@@ -311,13 +316,13 @@ class MetricsTab(TabWidget):
         mainLayout.addWidget(styleMapStyleLabel, 0, 4)
         mainLayout.addWidget(self.styleMapStyleDrop, 0, 5)
 
-        unitsPerEmLabel = QLabel("Units per em:", self)
-        ascenderLabel = QLabel("Ascender:", self)
-        capHeightLabel = QLabel("Cap height:", self)
-        italicAngleLabel = QLabel("Italic angle:", self)
-        descenderLabel = QLabel("Descender:", self)
-        xHeightLabel = QLabel("x-height:", self)
-        noteLabel = QLabel("Note:", self)
+        unitsPerEmLabel = QLabel(self.tr("Units per em:"), self)
+        ascenderLabel = QLabel(self.tr("Ascender:"), self)
+        capHeightLabel = QLabel(self.tr("Cap height:"), self)
+        italicAngleLabel = QLabel(self.tr("Italic angle:"), self)
+        descenderLabel = QLabel(self.tr("Descender:"), self)
+        xHeightLabel = QLabel(self.tr("x-height:"), self)
+        noteLabel = QLabel(self.tr("Note:"), self)
         # In the UFO specs these are integer or float, and unitsPerEm is
         # non-negative integer or float
         self.loadPositiveIntegerFloat(font, "unitsPerEm", "unitsPerEm")
@@ -376,21 +381,24 @@ class MetricsTab(TabWidget):
 class OpenTypeTab(TabWidget):
 
     def __init__(self, font, parent=None):
-        super().__init__(parent, name="OpenType")
+        super().__init__(parent)
+        self.name = self.tr("OpenType")
 
-        nameGroup = QGroupBox("name table", self)
+        nameGroup = QGroupBox(self.tr("name table"), self)
         # nameGroup.setFlat(True)
         nameLayout = QGridLayout(self)
 
-        preferredFamilyNameLabel = QLabel("Pref. Family Name:", self)
-        preferredSubfamilyNameLabel = QLabel("Pref. Subfamily Name:", self)
-        compatibleFullNameLabel = QLabel("Compatible Full Name:", self)
-        WWSFamilyNameLabel = QLabel("WWS Family Name:", self)
-        WWSSubfamilyNameLabel = QLabel("WWS Subfamily Name:", self)
-        versionLabel = QLabel("Version:", self)
-        uniqueIDLabel = QLabel("Unique ID:", self)
-        descriptionLabel = QLabel("Description:", self)
-        sampleTextLabel = QLabel("Sample text:", self)
+        preferredFamilyNameLabel = QLabel(self.tr("Pref. Family Name:"), self)
+        preferredSubfamilyNameLabel = QLabel(self.tr("Pref. Subfamily Name:"),
+                                             self)
+        compatibleFullNameLabel = QLabel(self.tr("Compatible Full Name:"),
+                                         self)
+        WWSFamilyNameLabel = QLabel(self.tr("WWS Family Name:"), self)
+        WWSSubfamilyNameLabel = QLabel(self.tr("WWS Subfamily Name:"), self)
+        versionLabel = QLabel(self.tr("Version:"), self)
+        uniqueIDLabel = QLabel(self.tr("Unique ID:"), self)
+        descriptionLabel = QLabel(self.tr("Description:"), self)
+        sampleTextLabel = QLabel(self.tr("Sample text:"), self)
         self.loadString(
             font, "openTypeNamePreferredFamilyName", "preferredFamilyName")
         self.loadString(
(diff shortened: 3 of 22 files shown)


Lib/defconQt/featureTextEditor.py
```diff
@@ -17,11 +17,11 @@ def __init__(self, font=None, parent=None):
         self.editor = FeatureTextEditor(self.font.features.text, self)
         self.resize(600, 500)
 
-        fileMenu = QMenu("&File", self)
-        fileMenu.addAction("&Save…", self.save, QKeySequence.Save)
+        fileMenu = QMenu(self.tr("&File"), self)
+        fileMenu.addAction(self.tr("&Save") + "…", self.save, QKeySequence.Save)
```
adrientetar: No need to split this

madig: You mean the ellipsis? Sadly there is a need: "pylupdate5 gets confused when strings contain non-ASCII characters". Unless you know a way to give the tool the correct slap on the back of its' head?

Lib/defconQt/featureTextEditor.py
```diff
@@ -17,11 +17,11 @@ def __init__(self, font=None, parent=None):
...
         fileMenu.addSeparator()
-        fileMenu.addAction("&Reload From Disk", self.reload)
-        fileMenu.addAction("E&xit", self.close, QKeySequence.Quit)
+        fileMenu.addAction(self.tr("&Reload From Disk"), self.reload)
+        fileMenu.addAction(self.tr("&Close"), self.close, QKeySequence.Quit)
```
adrientetar: On Windows it's usually Exit. Time for platform-specific string, I presume.

Lib/defconQt/featureTextEditor.py
```diff
@@ -29,6 +29,7 @@ def __init__(self, font=None, parent=None):
         self.editor.modificationChanged.connect(self.setWindowModified)
 
     def setWindowTitle(self, title="Font Features", font=None):
+        title = self.tr("Font Features")
```
adrientetar: title = self.tr(title)

Lib/defconQt/fontInfo.py
```diff
@@ -790,14 +785,14 @@ class PostScriptTab(TabWidget):
     def __init__(self, font, parent=None):
         super().__init__(parent, name="PostScript")
```
adrientetar:
```
name = self.tr("PostScript")
super().__init__(parent, name=name)
```

and similar for other tabs, then bring back InfoTabWidget.

Lib/defconQt/featureTextEditor.py
madig: Confused as in, it finds the strings and extracts them correctly, but on a subsequent run, it puts the string "Save?" insead of "Save…" into the translation file, invalidating its' translations.

Lib/defconQt/fontView.py
```diff
@@ -99,8 +99,7 @@ def loadGlyphList(self):
             try:
                 glyphList = glyphList.parseGlyphList(glyphListPath)
             except Exception as e:
-                msg = "The glyph list at %s cannot be parsed \
-                       and will be dropped." % glyphListPath
+                msg = self.tr("The glyph list at %s cannot be parsed and will be dropped.").format(glyphListPath)
```
adrientetar: 79-chars per line max. (should be highlighted in the Travis report).

Lib/defconQt/fontView.py
```diff
@@ -537,13 +536,13 @@ class AddGlyphsDialog(QDialog):
     def __init__(self, currentGlyphs=None, parent=None):
         super(AddGlyphsDialog, self).__init__(parent)
         self.setWindowModality(Qt.WindowModal)
-        self.setWindowTitle("Add Glyphs…")
+        self.setWindowTitle(self.tr("Add Glyphs") + "…")
```
adrientetar: No need to split

Lib/defconQt/fontView.py
```diff
@@ -537,13 +536,13 @@ class AddGlyphsDialog(QDialog):
...
         self.currentGlyphs = currentGlyphs
         self.currentGlyphNames = [glyph.name for glyph in currentGlyphs]
 
         layout = QGridLayout(self)
         self.importCharDrop = QComboBox(self)
-        self.importCharDrop.addItem("Import glyphnames…")
+        self.importCharDrop.addItem(self.tr("Import glyphnames") + "…")
```
adrientetar: No need to split

Lib/defconQt/fontView.py
```diff
@@ -615,20 +614,19 @@ class SortDialog(QDialog):
     def __init__(self, desc=None, parent=None):
         super(SortDialog, self).__init__(parent)
         self.setWindowModality(Qt.WindowModal)
-        self.setWindowTitle("Sort…")
+        self.setWindowTitle(self.tr("Sort") + "…")
```
adrientetar: idem

Lib/defconQt/fontView.py
```diff
@@ -615,20 +614,19 @@ class SortDialog(QDialog):
...
+            self.tr("A combination of simple, complex and custom sorts that give optimized ordering results."))
+        self.glyphSetBox = QRadioButton(self.tr("Glyph set"), self)
         self.glyphSetBox.toggled.connect(self.glyphSetToggle)
         self.glyphSetDrop = QComboBox(self)
         self.glyphSetDrop.setEnabled(False)
         glyphSets = readGlyphSets()
         for name, glyphNames in glyphSets.items():
             self.glyphSetDrop.addItem(name, glyphNames)
-        self.customSortBox = QRadioButton("Custom…", self)
+        self.customSortBox = QRadioButton(self.tr("Custom") + "…", self)
```
adrientetar: idem

Lib/defconQt/featureTextEditor.py
madig: Can't use variable names in translation marks.

Lib/defconQt/fontView.py
```diff
@@ -782,11 +780,11 @@ def __init__(self, font):
 
         menuBar = self.menuBar()
         # TODO: make shortcuts platform-independent
-        fileMenu = QMenu("&File", self)
-        fileMenu.addAction("&New…", self.newFile, QKeySequence.New)
-        fileMenu.addAction("&Open…", self.openFile, QKeySequence.Open)
+        fileMenu = QMenu(self.tr("&File"), self)
+        fileMenu.addAction(self.tr("&New") + "…", self.newFile, QKeySequence.New)
```
adrientetar: idem, & same for other similar occurences

Lib/defconQt/fontView.py
```diff
@@ -1519,6 +1517,7 @@ def __init__(self, parent=None):
         self.setLayout(mainLayout)
 
     def addGlyphSet(self, glyphNames=[], glyphSetName="New glyph set"):
+        glyphSetName = self.tr("New glyph set")
```
adrientetar: glyphSetName = self.tr(glyphSetName)

Lib/defconQt/featureTextEditor.py
madig: The naming is a bit unfortunate. "File" -> "Exit" suggests that it exits the program, or at least the file. "File" -> "Close" is still unfortunate, but a bit better imo.

Lib/defconQt/fontView.py
madig: See my comment above ;)

Lib/defconQt/fontView.py
madig: See my comment above ;)

Lib/defconQt/featureTextEditor.py
adrientetar: Okay let's keep this, but there are other windows in the program using Exit iirc so please change that too.

Lib/defconQt/featureTextEditor.py
adrientetar: Alright then remove the intelligence in setWindowTitle, make it

```
self.setWindowTitle(self.tr("Font Features") + " – %s %s" % (
              self.font.info.familyName, self.font.info.styleName))
```

in the caller and just prepend `[*]` in setWindowTitle before calling superclass.

Lib/defconQt/util/drawing.py
```diff
@@ -168,11 +168,11 @@ def drawFontVerticalMetrics(painter, glyph, scale, rect, drawLines=True,
...
-        ("Descender", font.info.descender),
-        ("Baseline", 0),
-        ("x-height", font.info.xHeight),
-        ("Cap height", font.info.capHeight),
-        ("Ascender", font.info.ascender),
+        (QCoreApplication.translate("GlyphView", "Descender"), font.info.descender),
+        (QCoreApplication.translate("GlyphView", "Baseline"), 0),
+        (QCoreApplication.translate("GlyphView", "x-height"), font.info.xHeight),
+        (QCoreApplication.translate("GlyphView", "Cap height"), font.info.capHeight),
+        (QCoreApplication.translate("GlyphView", "Ascender"), font.info.ascender),
```
adrientetar: drawing, not GlyphView

Lib/defconQt/fontInfo.py
```diff
@@ -38,7 +32,7 @@ def __init__(self, font, parent=None):
         mainLayout.addWidget(buttonBox)
         self.setLayout(mainLayout)
 
-        self.setWindowTitle("Font Info – %s %s" % (
+        self.setWindowTitle(self.tr("Font Info") + " – %s %s" % (
             self.font.info.familyName, self.font.info.styleName))
```
adrientetar: Can you avoid using + in your changes and use the existing % or .format for this purpose instead?

Lib/defconQt/fontView.py
adrientetar: Okay but then you need to stop overriding the function parameter, there's no point providing a parameter if it becomes a constant value.

````
[...] glyphSetName=None):
if glyphSetName is None:
    glyphSetName = self.tr("New glyph set")

File: Lib/defconQt/fontInfo.py (hunk 13 above)
Comment (madig) [inline Lib/defconQt/fontInfo.py]: Okay. What about the ellipsiseseses? Replace by "\u2026"? Ugly for translators but would get around `pylupdate5`'s clumsiness.

Comment (adrientetar) [inline Lib/defconQt/fontInfo.py]: I asked the question on the mailing list.

File: Lib/defconQt/fontInfo.py (hunk 4 above)
Comment (madig) [inline Lib/defconQt/fontInfo.py]: I tried this at first, but get `RuntimeError: super-class __init__() of type GeneralTab was never called`.

File: Lib/defconQt/featureTextEditor.py (hunk 1 above)
Comment (khaledhosny) [inline Lib/defconQt/featureTextEditor.py]: It is a bad l10n practice in general to split translatable strings, you don’t know what languages will want to do with the string (reorder it, add spaces, use different punctuation, and so on).

File: Lib/defconQt/featureTextEditor.py (hunk 3 above)
Comment (khaledhosny) [inline Lib/defconQt/featureTextEditor.py]: This should be a single string, not split into translatable and non-translatable parts.

Comment (khaledhosny) [inline Lib/defconQt/featureTextEditor.py]: And please use any of the newer Python formatting features, the `printf` style formatting in Python is not good for l10n.

File: Lib/defconQt/fontView.py [hunk 14]
```diff
@@ -782,11 +784,12 @@ def __init__(self, font):
...
-        fileMenu = QMenu("&File", self)
-        fileMenu.addAction("&New…", self.newFile, QKeySequence.New)
-        fileMenu.addAction("&Open…", self.openFile, QKeySequence.Open)
+        fileMenu = QMenu(self.tr("&File"), self)
+        fileMenu.addAction(self.tr("&New…"), self.newFile, QKeySequence.New)
+        fileMenu.addAction(self.tr("&Open…"), self.openFile,
+                           QKeySequence.Open)
         # recent files
-        self.recentFilesMenu = QMenu("Open &Recent…", self)
+        self.recentFilesMenu = QMenu(self.tr("Open &Recent"), self)
```
khaledhosny: The ellipsis is lost here.

Lib/defconQt/featureTextEditor.py
madig: Alright. The newest patches reunite the strings :) Now only pylupdate5 needs to work correctly.

Lib/defconQt/fontView.py
```diff
@@ -907,8 +920,8 @@ def saveFile(self, path=None, ufoFormatVersion=3):
 
     def saveFileAs(self):
         fileFormats = OrderedDict([
-            ("UFO Font version 3 (*.ufo)", 3),
-            ("UFO Font version 2 (*.ufo)", 2),
+            (self.tr("UFO Font version 3") + " (*.ufo)", 3),
+            (self.tr("UFO Font version 2") + " (*.ufo)", 2),
         ])
```
khaledhosny: This should use `"".format()` like the cases below.

Lib/defconQt/fontView.py
madig: I forgot to mention this in the commit message, I removed it on purpose because it does not require additional user input. Gimp doesn't have an ellipsis there either.

Lib/defconQt/fontView.py
```diff
@@ -961,13 +976,14 @@ def exportFile(self):
             if getattr(self.font.info, attr) is None:
                 missingAttrs.append(attr)
         if len(missingAttrs):
-            text = "Attributes required for font generation are missing: %s" \
-                   % " ".join(missingAttrs)
+            text = self.tr("Attributes required for font generation are "
+                           "missing: %s") % " ".join(missingAttrs)
```
khaledhosny: Use `"".format()` here please as well.

Lib/defconQt/fontView.py
```diff
@@ -1064,15 +1080,15 @@ def maybeSaveBeforeExit(self):
                 # TODO: maybe cache this font name in the Font
                 currentFont = os.path.basename(self.font.path.rstrip(os.sep))
             else:
-                currentFont = "Untitled.ufo"
-            body = "Do you want to save the changes you made to “%s”?" \
-                % currentFont
+                currentFont = self.tr("Untitled.ufo")
+            body = self.tr("Do you want to save the changes you made "
+                           "to “%s”?") % currentFont
```
khaledhosny: Use `"".format()` here please as well.

Lib/defconQt/fontInfo.py
madig: I found a workaround: set name after super-init. I'm so clever.

Lib/defconQt/fontView.py
```diff
@@ -1243,7 +1259,7 @@ def _selectionChanged(self, selection):
                 count = selection
                 text = ""
             if not count == 0:
-                text = "%s(%d selected)" % (text, count)
+                text = text + self.tr("({0} selected)").format(count)
```
khaledhosny: This will need to use plural forms.

Lib/defconQt/fontView.py
```diff
@@ -1584,11 +1605,11 @@ def writeGlyphSets(self, settings):
 
     def getGlyphList(self):
         fileFormats = (
-            "Text file (*.txt)",
-            "All files (*.*)",
+            self.tr("Text file") + " (*.txt)",
+            self.tr("All files") + " (*.*)",
```
khaledhosny: Use `"".format()` here please as well.

Lib/defconQt/fontView.py
```diff
@@ -1740,11 +1763,11 @@ def __init__(self, parent=None):
     def addItem(self):
 
         def mangleNewName():
-            name = "New"
+            name = self.tr("New")
             index = 0
             while self.markColorWidget.findItems(name, Qt.MatchExactly, 1):
                 index += 1
-                name = "New ({})".format(index)
+                name = name + " ({})".format(index)
```
khaledhosny: The concentrated strings here are bad for l10n, this needs to be an explicit list of complete `New foo` strings.

Lib/defconQt/metricsWindow.py
```diff
@@ -43,8 +43,8 @@ def __init__(self, font, string=None, pointSize=defaultPointSize,
             try:
                 string = getuser()
             except:
-                string = "World"
-            string = "Hello %s" % string
+                string = self.tr("World")
+            string = self.tr("Hello") + " " + string
```
khaledhosny: The concatenation here needs to be avoided, too.

Lib/defconQt/scriptingWindow.py
```diff
@@ -15,14 +15,14 @@ def __init__(self):
...
+        fileMenu = QMenu(self.tr("&File"), self)
+        fileMenu.addAction(self.tr("&Run…"), self.runScript, "Ctrl+R")
         fileMenu.addSeparator()
-        fileMenu.addAction("E&xit", self.close, QKeySequence.Quit)
+        fileMenu.addAction(self.tr("&Close"), self.close, QKeySequence.Quit)
         self.menuBar().addMenu(fileMenu)
 
         self.setCentralWidget(self.editor)
-        self.setWindowTitle("[*]Untitled.py")
+        self.setWindowTitle(self.tr("[*]Untitled.py"))
```
khaledhosny: I think this should be translatable if the file name is hardcoded.

Lib/defconQt/fontView.py
madig: I have no idea how to mark stuff for pluralizing in Python...

Lib/defconQt/scriptingWindow.py
madig: It is?

Lib/defconQt/featureTextEditor.py
madig: I think I got them all. By the way, metricsWindow.py features a setupFileMenu() that is never called. Purpose?

madig:
@adrientetar want me to open a fresh branch with one big clean patch and without the translation files?

Lib/defconQt/__main__.py
```diff
@@ -1,7 +1,9 @@
 from defconQt import __version__, representationFactories
 from defconQt import icons_db  # noqa
 from defconQt.fontView import Application
-from PyQt5.QtCore import QCommandLineParser, QSettings
+from PyQt5.QtCore import (QCommandLineParser, QSettings,
+                          QCoreApplication, QTranslator, QLocale,
+                          QLibraryInfo)
```
adrientetar: Can you make a newline after `from PyQt5.QtCore import (` rather?

Lib/defconQt/__main__.py
```diff
@@ -17,13 +19,31 @@ def main():
...
+                       "/resources")
+    app.installTranslator(appTranslator)
+
     app.loadGlyphList()
     # parse options and open fonts
     parser = QCommandLineParser()
-    parser.setApplicationDescription("The TruFont font editor.")
+    parser.setApplicationDescription(
+        QCoreApplication.translate("Command-line parser",
+                                   "The TruFont font editor."))
```
adrientetar: newline after `parser.setApplicationDescription(QCoreApplication.translate(`

Lib/defconQt/__main__.py
```diff
@@ -17,13 +19,31 @@ def main():
...
     parser.addHelpOption()
     parser.addVersionOption()
-    parser.addPositionalArgument("files", "The UFO files to open.")
+    parser.addPositionalArgument(
+        QCoreApplication.translate("Command-line parser", "files"),
+        QCoreApplication.translate("Command-line parser",
+                                   "The UFO files to open."))
```
adrientetar: Same here.

Lib/defconQt/fontInfo.py
```diff
@@ -180,22 +180,24 @@ class GeneralTab(TabWidget):
 
     def __init__(self, font, parent=None):
         super().__init__(parent, name="General")
+        self.name = self.tr("General")
```
adrientetar: This duplication needs to go. Can you just remove the name arg from \__init__ call?

Lib/defconQt/fontInfo.py
```diff
@@ -285,14 +287,17 @@ class MetricsTab(TabWidget):
 
     def __init__(self, font, parent=None):
         super().__init__(parent, name="Metrics")
+        self.name = self.tr("Metrics")
```
adrientetar: Same here and for other tabs

Lib/defconQt/fontView.py
```diff
@@ -313,12 +314,13 @@ def __init__(self):
...
 
-        layerSetGroup = QGroupBox("Layers", self)
+        layerSetGroup = QGroupBox(self.tr("Layers"), self)
         layerSetGroup.setFlat(True)
         layerSetLayout = QVBoxLayout(self)
 
         self.layerSetWidget = QTreeWidget(self)
-        self.layerSetWidget.setHeaderLabels(("Layer Name", "Color"))
+        self.layerSetWidget.setHeaderLabels((self.tr("Layer Name"),
+                                             self.tr("Color")))
```
adrientetar: I rather make a new line after function opening brace here, when you make a new line after opening brace you can put just one more level of indent on the next line under PEP8.

Lib/defconQt/glyphCollectionView.py
```diff
@@ -30,11 +30,18 @@
...
 def proceedWithDeletion(self, erase=False):
-    do = "clear" if not erase else "delete"
+    clear_header = QCoreApplication.translate(
+        "GlyphCollectionWidget", "Clear glyphs")
+    clear_text = QCoreApplication.translate(
+        "GlyphCollectionWidget", "Do you want to clear selected glyphs?")
+    erase_header = QCoreApplication.translate(
+        "GlyphCollectionWidget", "Delete glyphs")
+    erase_text = QCoreApplication.translate(
+        "GlyphCollectionWidget", "Do you want to delete selected glyphs?")
```
adrientetar: tbh I think we could just make `proceedwithdeletion` a mthod of GlyphCollectionView and so use self.tr here.

Lib/defconQt/groupsView.py
```diff
@@ -184,6 +184,7 @@ def __init__(self, font, parent=None):
         self.setWindowTitle(font=self.font)
 
     def setWindowTitle(self, title="Groups Window", font=None):
+        title = self.tr("Groups Window")
```
adrientetar: Can you put title=None as default parameter and do if title is None: etc.?

Lib/defconQt/metricsWindow.py
```diff
@@ -41,10 +41,9 @@ def __init__(self, font, string=None, pointSize=defaultPointSize,
 
         if string is None:
             try:
-                string = getuser()
+                string = self.tr("Hello {0}").format(getuser())
```
adrientetar: Shouldn't .format be called on the result of self.tr?

Lib/defconQt/metricsWindow.py
```diff
@@ -77,12 +76,13 @@ def __init__(self, font, string=None, pointSize=defaultPointSize,
...
-        fileMenu = QMenu("&File", self)
-        fileMenu.addAction("&Save…", self.save, QKeySequence.Save)
-        fileMenu.addAction("E&xit", self.close, QKeySequence.Quit)
+        fileMenu = QMenu(self.tr("&File"), self)
+        fileMenu.addAction(self.tr("&Save…"), self.save, QKeySequence.Save)
+        fileMenu.addAction(self.tr("&Close"), self.close, QKeySequence.Quit)
         self.menuBar().addMenu(fileMenu)
 
     def setWindowTitle(self, title="Metrics Window", font=None):
+        title = self.tr("Metrics Window")
```
adrientetar: Same here with default parameter, switch it to None and change logic acordingly.

Lib/defconQt/objects/glyphDialogs.py
```diff
@@ -12,20 +12,23 @@ class GotoDialog(QDialog):
...
-        self.setWindowTitle("Go to…")
+        self.setWindowTitle(self.tr("Go to…"))
         self.font = currentGlyph.getParent()
         self._sortedGlyphs = self.font.unicodeData.sortGlyphNames(
             self.font.keys(), self.alphabetical)
 
         layout = QGridLayout(self)
-        self.glyphLabel = QLabel("Glyph:", self)
+        self.glyphLabel = QLabel(
+            QCoreApplication.translate("GotoDialog", "Glyph:"), self)
```
adrientetar: Did self.tr not work here?

adrientetar: (and below)

Lib/defconQt/util/platformSpecific.py
```diff
@@ -1,14 +1,15 @@
-from PyQt5.QtCore import Qt
+from PyQt5.QtCore import Qt, QCoreApplication
 import sys
 
 # -----------
 # File dialog
 # -----------
+ufoFileFormat = QCoreApplication.translate("File dialog", "UFO Fonts {}")
```
adrientetar: Can you put prefix _ to this variable? It's not meant to be exposed.

adrientetar:
Yes this PR could use some squashes in the commit log. Also could you use QApplication instead of QCoreApplication? QApplication inherit the former and that's what we usually import so we don't need QCoreApplication explicitly here (it's for applications without GUIs so it's a parent of QApplication).

Good job! Fix the latest comments and this should be good to go. Thanks.

Lib/defconQt/glyphCollectionView.py
madig: My line of thinking. Was there a reason to make it a stand-alone function?

adrientetar: Genericity. But it's not needed anywhere else, afaict.

Lib/defconQt/metricsWindow.py
madig: Uhm, it is?

adrientetar: Right, I misread. Sorry.

Lib/defconQt/objects/glyphDialogs.py
madig: The goto dialog gets repurposed by that other dialog, whatsitcalled, which mucks with self.tr. Translations would appear in the goto dialog but not in the other.

adrientetar: OK.

Lib/defconQt/fontView.py
```diff
@@ -1243,7 +1259,7 @@ def _selectionChanged(self, selection):
                 count = selection
                 text = ""
             if not count == 0:
-                text = "%s(%d selected)" % (text, count)
+                text = self.tr("{0}({1} selected)").format(text, count)
```
khaledhosny: This should be something like `self.tr("{0}({1} selected)".format(text, count), n=count)` to use plural forms.

madig: Hm, it won't fail, but I also see nothing in Linguist indicating that I can pluralize...

madig: Ah, I found it!

Lib/defconQt/glyphCollectionView.py
madig: There's a joke about computer scientists in here somewhere :grin:

madig: Will be an extra commit.

madig:
Squashed, without translations.

Lib/defconQt/featureTextEditor.py
adrientetar: Can be removed, thanks.

adrientetar:
Can you remove setupFileMenu in metricsWindow, and squash it all?

adrientetar:
I can merge as-is otherwise.

adrientetar:
Thanks!

madig:
Wah, wait, i didn't delete the code yet :D

PR #229, opened 2016-02-01
Font Info: Align labels right

madig:
A first sweep. Still needs a more pleasant resizing behavior.

Other dialogs will follow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ src/
 __pycache__/
 *.pyc
 *.spec
+.*.swp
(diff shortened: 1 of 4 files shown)


adrientetar:
Maybe we could subclass QLabel into RLabel class no there's no need to repeat?

madig:
Hm, might be a good idea. Will check the source in the coming days if it could be useful elsewhere.

davelab6:
I am contracted by Google to manage the collection.

I just wrote and published
https://github.com/googlefonts/fontbakery/pull/661/files so you can see
where I hope to go this year which will help Debian and similar folks,
since at the moment a lot of the families aren't developed like regular
(libre) software and probably can't belong in main just yet.

madig:
@davelab6 Wrong window?

davelab6:
lol woops. deleting :)

PR #233, opened 2016-02-06
Update README.md

kelvin13:
pip command will not work on Ubuntu

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ PyQt5.
 
 ## Getting started
 
-1. Install dependencies with `pip install -r requirements.txt`
+1. Install dependencies with `pip3 install -r requirements.txt`
 
 2. Install using `python setup.py install && trufont` or run under virtualenv:
    `cd Lib && python -m defconQt`


davelab6:
I suggest using `pip3` on line 17 of README too, and where it appears in INSTALL.git

adrientetar:
Because pip is Python 2 is you have Python 2 installed. There is no pip3 command on Windows.

anthrotype:
Alternatively you could write `{path/to/python3.x} -m pip install -r requirements.txt` to make sure one uses the pip which is associated with the right interpreter.

adrientetar:
Can we assume that Python 3 being in the requirements means people should actually use Python 3?

davelab6:
We can not assume that, clearly.

I would prefer pip3 (since we know Mac OS X is the most popular OS for this
application's audience and exists there, always alongside pip which ==
pip2) and mention for Windows that it is pip.

kelvin13:
Why is this so controversial just put “pip3” with a link to pip3 install instructions and underneath in parentheses put (windows users use “pip”)

anthrotype:
> There is no pip3 command on Windows.

I just checked on my windows virtual machine and `pip3.exe` (like `pip3.5.exe`) is actually present. Maybe @adrientetar got confused by the fact that on Windows usually there's no `python3.exe` executable, but only `python.exe`.

So I think it's fine to tell user to use `pip3` even if that should be implied. Maybe @kelvin13 could also update the other reference to `pip` command in the following paragraph, like @davelab6 suggested?

adrientetar:
I'll remove it it's redundant.

PR #235, opened 2016-02-07
[README.md] add install info on python 3, pyqt5 and virtual environment

anthrotype:
I was reading the README.md and I noticed that it says

> run under virtualenv: `cd Lib && python -m defconQt`

That's a bit misleading. The command simply runs trufont in place, it doesn't mean it runs from an isolated virtual environment, as the reference to `virtualenv` seems to imply.

I think it'd be better if we recommend users, and especially developers, to install trufont and its many dependencies inside a virtual environment (a "real" one ;).

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,34 @@ PyQt5.
 
 ## Getting started
 
-1. Install dependencies with `pip3 install -r requirements.txt`
+1. Install Python 3 
+    - OS X: `brew install python3 && brew linkapps python3`
+    - Windows: Download installer from https://www.python.org/downloads/
+    - Linux: It's usually packaged with the OS.
 
-2. Install using `python setup.py install && trufont` or run under virtualenv:
-   `cd Lib && python -m defconQt`
+2. Install PyQt5:
+    - OS X: `brew install pyqt5`
+    - Windows: You can download the binary installers (32 or 64 bit) from:
+        https://riverbankcomputing.com/software/pyqt/download5
+    - Linux: Follow instructions to compile PyQt5 (and SIP) from source:
+        http://pyqt.sourceforge.net/Docs/PyQt5/installation.html
+
+3. Set up a new virtual environment using `pyvenv` (or `virtualenv`), and give
+    it access to the system site-packages to make sure PyQt5 can be imported:
+        `pyvenv --system-site-packages trufont`
+    
+4. Activate the newly created environment:
+    - OS X or Linux: `source trufont/bin/activate`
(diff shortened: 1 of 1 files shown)


adrientetar:
virtualenv is one more step. What does it bring to the table?

anthrotype:
> virtualenv is one more step. What does it bring to the table?
I know but pyvenv does not install pip & setuptools when passing the `--system-site-packages` option for unknown reason. virtualenv does.
Also virtualenv supports more shells. E.g. I can use it in msysgit or msys2 on Windows with `source venv/Scripts/activate`, whereas pyvenv only gives you `activate.bat`.

anthrotype:
Also, since the dependencies are (except for setuptools) installed in editable mode from git repos, I think it would make sense to also install trufont itself in editable mode, so that when one `git pull` the latest sources, it is already up to date, without needing a further `pip install --upgrade .`

anthrotype:
Thanks!

README.md
```diff
@@ -7,10 +7,37 @@ PyQt5.
...
+    - Linux: Follow instructions to compile PyQt5 (and SIP) from source:
+        http://pyqt.sourceforge.net/Docs/PyQt5/installation.html
+
+3. Set up a new Python virtual environment using `virtualenv`.
+    - Install or update the virtualenv module with `pip3 install --upgrade virtualenv`.
+       You may require `sudo` access on Linux. Alternatively, you can install it in
+       the Python user directory: `pip3 install --user --upgrade virtualenv`.
+    - Create a new virtual environment, and give it access to the system
+        site-packages to make sure PyQt5 can be imported:
+        `python3 -m virtualenv --system-site-packages trufont`
```
anthrotype: I hope Windows people don't get back to us with "hey, I did install Python 3 but there's no python3 executable"...
The reason I didn't simply use the `virtualenv` command here, but called it with `python3 -m ...`, is that one may already have `virtualenv` installed in python2.7.

Virtualenv als has a  `--python` option to specify the name of the interpreter to use. This way it wouldn't matter if you use the py27 or py35 virtualenv, you would get what you ask. The problem is you still need to name the python3 interpreter and we can't call it `python3` as this name is not available on vanilla Python for Windows.

So either this or that

```diff
- python3 -m virtualenv --system-site-packages trufont
+ virtualenv --python=python3 --system-site-packages trufont
```

neither works on Windows, where python is simply `python.exe`.
One way they can make sure they are using the correct interpreter version is using the Python Launcher for Windows:

```
py -3 -m virtualenv ...
```

Anyway.

PR #248, opened 2016-02-25
Reworked Font Info dialog

madig:
First mostly-skeleton.

*Not for merging*, rather for comments. Most non-QLineEdits are dummies still and QCheckBoxes need to have their layout flipped.

There is a lot of repetition in the way the vast majority of fields in the Opentype and Postscript tab are set up, I want to do a compression pass after I get it all working.

### DIFF
--- a/Lib/defconQt/fontInfo.py
+++ b/Lib/defconQt/fontInfo.py
@@ -1,12 +1,15 @@
-from PyQt5.QtCore import QDate, QDateTime, QRegularExpression, QTime, Qt
+from PyQt5.QtCore import (
+    QDate, QDateTime, QRegularExpression, QTime, Qt)
 from PyQt5.QtGui import (
-    QDoubleValidator, QIntValidator, QRegularExpressionValidator,
-    QStandardItem, QStandardItemModel)
+    QRegularExpressionValidator, QStandardItem, QStandardItemModel)
 from PyQt5.QtWidgets import (
-    QCheckBox, QComboBox, QDateTimeEdit, QDialog, QDialogButtonBox,
-    QGridLayout, QGroupBox, QLabel, QLineEdit, QListView, QPlainTextEdit,
-    QTabWidget, QVBoxLayout, QWidget)
-from defconQt.util.rlabel import RLabel
+    QCheckBox, QComboBox, QDateTimeEdit, QDialog, QDialogButtonBox, QGroupBox,
+    QLabel, QLineEdit, QListView, QPlainTextEdit, QTabWidget, QVBoxLayout,
+    QHBoxLayout, QFormLayout, QWidget, QScrollArea, QSpacerItem, QSizePolicy,
+    QSpinBox)
+from defconQt.util.rcheckbox import RCheckBox
+from defconQt.util.bitlistview import BitListView
+from defconQt.util.floatspinbox import FloatSpinBox
 
 
 class InfoTabWidget(QTabWidget):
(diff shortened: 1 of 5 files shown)


adrientetar:
> generalLayout = QHBoxLayout()

QHBoxLayout(self) & same for other layouts.

> self.nameRecordsEdit = QTableWidget()

QTableWidget(self) etc.

madig:
Is that necessary when I add them to layouts a few lines later?

adrientetar:
Well I do this for every widget so it alerts me when absent when reading the code.

madig:
Doing `QHBoxLayout(self)` for sub-layouts gives me warnings like `QLayout: Attempting to add QLayout "" to GeneralTab "", which already has a layout`. So it works for the top-level layout but not for everything below it. Will try with widgets.

adrientetar:
> Doing QHBoxLayout(self) for sub-layouts gives me warnings like QLayout: Attempting to add QLayout "" to GeneralTab "", which already has a layout.

Right. I don't do much nested layouts, my bad.

adrientetar:
I think we can remove RLabel definition altogether given QFormLayout.

madig:
On my todo list, it's still in use by the inspector window.

madig:
Is there a specific reason integer/float attributes are presented as a QLineEdit?

adrientetar:
Instead of?

madig:
QSpinBox, QDoubleSpinBox

adrientetar:
Isn't that what QIntValidator/QDoubleValidator are already doing? Where were you allowed to type invalid input?

madig:
Yes, but QSpinBox does validation, probably also correct decimal mark handling per locale and gives the user arrows and up/down arrow handling :) I find it a more natural fit.

By invalid input I mean things that pass through QIntValidator/QDoubleValidator but fail validation by ufolib upon assignment.

madig:
Found it. `openTypeHeadCreated`. Enter something bogus and you get a backtrace or a crash. Guess I'll make it a QDateTimeEdit thing.

madig:
It just occurred to me that maybe the dialog should only save attributes that the user changed, otherwise Trufont thinks the UFO has changed just because the user opened the font info dialog and clicked Ok without changing anything. This is getting more complicated. Ugh.

adrientetar:
Really? Normally setters in ufo3 defcon return early when given the same value. If not it should be investigated.

adrientetar:
Oh nevermind I'm not sure it does that for font info attributes since it's dict object. I'll check tomorrow.

madig:
Thanks. I think the dialog is almost done. Comments please :) If the code is okay, I'll take care of the unit test.

madig:
Hm. Fields without a checkbox (General, Legal) are saved as empty strings and turn up as such in fontinfo.plist. Should those empty strings be None'd? Empty edit boxes with checkboxes should save empty strings imo, because a checkmark implies the user wanted an empty string.

madig:
Uh wait, the code is not ready. I forgot I still have to implement the gasp and name record table thingies. Gah.

madig:
@adrientetar in the interest of finally getting this in, how about I remove the gasp and name record tables for now and fix the unit tests, then work on the omitted tables at a later point? I'm still trying to wrap my head around the model/view thing and it will probably take some more time...

adrientetar:
Sure.

madig:
@adrientetar Done. Ready for comments and merging :)

madig:
Huh. Why is the test failing.

davelab6:
> sudo apt-get install -qq -y python3-pyqt5 python3-pip python3-flake8

I don't see why this would timeout...

khaledhosny:
Just restart the build and see, it happened to me more than once that apt-get commands timed out on Travis, slow network or somesuch.

madig:
If in doubt, try turning it off and on again :)

adrientetar:
Question: why, e.g. for the General tab are there multiple QFormLayout used? It wouldn't be possible to fit the multiple areas on one form layout?

madig:
I haven't found a way to nest or subdivide QFormLayouts, so I nested them inside box layouts instead. This sadly means that fields do not align over the entire tab in e.g. OpenType and PostScript.

adrientetar:
Landed, thanks for this great work.

PR #253, opened 2016-03-19
Font info

adrientetar:
Closes #248.


madig:
Nooo, my links to the spec :(

What range can postscriptUniqueID take? The CFF spec just says it's an "integer".

adrientetar:
> What range can postscriptUniqueID take?

UniqueID [doesn't matter](https://fontforge.github.io/UniqueID.html) very much and tbh I don't really know what range hw from the 80s can take &ndash; [this document](http://partners.adobe.com/public/developer/en/font/5014.CIDFont_Spec.pdf) mentions 16,777,215.

Issue #240, opened 2016-02-09
Zoom in Glyph view


brawer:
- [ ] It would be nice if one could zoom into details in the glyph view. Maybe like this:

![trufontzoom](https://cloud.githubusercontent.com/assets/1527880/12930658/e4afb148-cf79-11e5-805d-ce2a8fbf50ee.png)
- [X] On MacOS, ⌘+ would zoom in; ⌘– would zoom out. Other platforms probably have similar conventions.
- [ ] By the way, the Glyphs font editor shows a nice grid at 10000 pt and above, with all editing snapped to the grid. Really convenient for editing.

<img width="458" alt="screen shot 2016-02-09 at 10 14 50 pm" src="https://cloud.githubusercontent.com/assets/1527880/12930827/d870f422-cf7a-11e5-95ad-c2620582e2ba.png">

<img width="465" alt="screen shot 2016-02-09 at 10 15 47 pm" src="https://cloud.githubusercontent.com/assets/1527880/12930841/e381a334-cf7a-11e5-8e62-ad33b1854890.png">

anthrotype:
BTW, is there any way to zoom in/out in the current glyph view? I could look at the source code but I'm lazy.

adrientetar:
It's Ctrl + wheel currently.

adrientetar:
I'll add +/-.

anthrotype:
Thanks!

davelab6:
Yes, hotkeys - and = for out and in would be great :)

anthrotype:
"z" and "x" hotkeys to zoom in/out would also be nice. I think both FontLab and Robofont use those. Maybe it could be user-configurable somehow.

anthrotype:
btw, also FontForge uses "z" and "x"

pathumego:
Editable shortcuts would be nice for all :)

adrientetar:
8c063ed

anthrotype:
I'd really like to have "z" and "x" shortcut for zoom in/out, like FontLab, Robofont and FontForge also do.
I could give this a go in the spare time. Any suggestions as to where I should be looking at is welcome.

adrientetar:
e870842c1f93909fbf532358bd9330a02fa80965

anthrotype:
Thanks!!! 🎉

davelab6:
Love it :)

PR #256, opened 2016-04-14
Rewrite most things

### LINKED ISSUE #254: [bezierMath] lineDistance raises "TypeError: 'NoneType' object is not iterable"
`bezierMath.lineProjection` returns None if the point cannot be projected on the segment. However `bezierMath.lineDistance` that calls `lineProjection` does not check if the returned value is None, and therefore makes Python crash with `TypeError: 'NoneType' object is not iterable`.

Calling `lineProjection` with  `ditchOutOfSegment=False` prevents it to return None and fixes the issue.
`ditchOutOfSegment` argument is True by default.

adrientetar:
cc #240.
Closes #254.


davelab6:
AWESOME

davelab6:
"1 check failed"?

moyogo:
@davelab6 no worries, we’ll clean that up

adrientetar:
@moyogo Thanks Denis !!

davelab6:
:D

PR #265, opened 2016-04-16
travis: run tests on OS X instead of Linux

anthrotype:
pyqt5 on Ubuntu Trusty is too old; we shall re-enable it once we figure out how to install an updated pyqt5 (>=5.5.1) without having to compile it from source...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,12 @@ sudo: false
 language: generic
 os: osx
 before_install:
-  # use pyenv to install python 3.5.1
-  - git clone https://github.com/yyuu/pyenv.git ~/.pyenv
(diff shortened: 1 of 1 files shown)


adrientetar:
There may be a way to remove usage of QProxyStyle, I'll look into that.

anthrotype:
@adrientetar the tests now pass on the Travis OSX machine.

anthrotype:
BTW, I just found out that pytest to help write tests for PyQt: https://pytest-qt.readthedocs.org/
we shall look into that as well.

adrientetar:
Sounds good. Thanks for this PR!

PR #270, opened 2016-04-18
README.md: update install instructions for pyqt5 on Win and OSX

anthrotype:
also, use a different name for example virtual environment, as 'trufont' can be confusing.

@pathumego, looks good?


pathumego:
👍

davelab6:
Great!

PR #281, opened 2016-04-28
travis: run tests on both OS X and Ubuntu Trusty

anthrotype:
fingers crossed...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,14 @@
-sudo: false
-language: generic
-os: osx
+matrix:
+  include:
+    - sudo: required
+      dist: trusty
+    - language: generic
+      os: osx
+
 before_install:
-  # use homebrew to install python 3.5.1
-  - brew update || brew update
-  - if test $(brew list | grep -q 'python3'); then brew upgrade python3; else brew install python3; fi
+  # on OSX, we use homebrew to install python 3.5.1
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then brew update || brew update; fi
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then if test $(brew list | grep -q 'python3'); then brew upgrade python3; else brew install python3; fi; fi
   - python3.5 -m pip install --user --upgrade pip virtualenv
   # create virtual environment
   - python3.5 -m virtualenv .env
@@ -17,6 +21,12 @@ before_install:
   - pip install -r requirements.txt
 install:
   - pip install -v .
+before_script:
+  # on Linux, we run "headless" tests with xvfb (X Virtual Framebuffer)
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then export DISPLAY=:99.0; fi
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sh -e /etc/init.d/xvfb start; fi
+  # give xvfb some time to start
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then sleep 3; fi 
 script:
   - flake8 --config=setup.cfg Lib/trufont
   - py.test -v -r a tests/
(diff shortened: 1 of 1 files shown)


adrientetar:
I think you need to upgrade pyenv on trusty though.

anthrotype:
yeah. such a pain

anthrotype:
I don't understand... I'm now getting a "core dumped" when running the test on Trusty Python 3.5.0:

https://travis-ci.org/trufont/trufont/jobs/126435518#L495

Even weirder, if I run locally on my 14.04 virtual machine, the `py.test` command passes successfully, but if I then try to launch `trufont` I can't see anything and in the console I get this error:

```
$ trufont

(python3.5:14972): Gtk-CRITICAL **: IA__gtk_widget_style_get: assertion 'GTK_IS_WIDGET (widget)' failed

QWidget::setMinimumSize: (/FontWindow) The largest allowed size is (16777215,16777215)

QXcbShmImage: shmget() failed (22: Invalid argument) for size 1073610756 (16383x16383)
```

adrientetar:
Right, iirc there's a bug with Qt GTK backend and QMainWindow sometimes on linux

https://bugs.launchpad.net/ubuntu/+source/subdownloader/+bug/832340

adrientetar:
What if you try to change style? It seems the GTK style is what can cause this:

http://stackoverflow.com/a/34249929

anthrotype:
Hm..
I got `trufont` to run with this little change, but I'm not sure it's the right thing to do:

```diff
diff --git a/Lib/trufont/__main__.py b/Lib/trufont/__main__.py
index 57b4798..21964f2 100644
--- a/Lib/trufont/__main__.py
+++ b/Lib/trufont/__main__.py
@@ -21,6 +21,7 @@ def main():
     representationFactories.registerAllFactories()
     # initialize the app
     app = Application(sys.argv)
+    app.setStyle("Fusion")
     app.setOrganizationName("TruFont")
     app.setOrganizationDomain("trufont.github.io")
     app.setApplicationName("TruFont")
```

anthrotype:
I got the tip from the last answer here:
http://stackoverflow.com/questions/14606396/gtk-critical-ia-gtk-widget-style-get-assertion-gtk-is-widget-widget-fa

adrientetar:
You can use the -style command line switch to change style, e.g. -style Plastique

anthrotype:
yes, also the `-style` argument works.

but still this does not explain why on Travis we get "core dumped", while locally at least we can get it running, though by disabling the GTK style...

adrientetar:
Someone reported the bug to PyQt once but Phil couldn't repro apparently: https://riverbankcomputing.com/pipermail/pyqt/2015-April/035820.html

adrientetar:
It seems there are patches for this issue on the Debian packages: https://bugs.launchpad.net/ubuntu/+source/qt4-x11/+bug/805303

Also off the top of my head there is a problem that has to do with overlay scrollbars (on Ubuntu Unity the scrollbars aren't always drawn but paint over widget when scrolling), this can be fixed by exporting LIBOVERLAY_SCROLLBAR=0 before starting the app, see https://bugs.launchpad.net/overlay-scrollbar/+bug/1014631  
Otherwise there is a patch that fixes those things (cf. first link) but I suspect it's patched only in debian packages and not the mainline Qt:

anthrotype:
I think i'm off for today.. If you wish to apply the patches, then tomorrow I can try rebasing on master and see if we get lucky. Sorry :-/

adrientetar:
I tried setting LIBOVERLAY_SCROLLBAR=0 but the build still fails, tho I'm pretty sure doing this fixed the crash on boot on Ubuntu back when I tried running it.

anthrotype:
I eventually got it to work by installing `libegl1-mesa` package from apt-get. Somehome, Qt could not find the `libEGL.so.1` library, which was needed in order to load `plugins/platforms/libqxcb.so`.
I'll squash and clean it up tomorrow. Glad we finally made it to work :)

adrientetar:
Wew, great job with this :shit: !

anthrotype:
squashed. we can merge it after the build passes 💃

adrientetar:
:+1:

PR #284, opened 2016-04-30
Draw icons in Hi-DPI mode

khaledhosny:
Set the AA_UseHighDpiPixmaps so that Qt returns larger pixmaps for
Hi-DPI systems.

Before:

After:


adrientetar:
I'm surprised this isn't enabled by default.

PR #299, opened 2016-07-18
Select anchors and components with “Select All”

khaledhosny:
Currently it does not select anchors at all, and selects components only
when the glyph has no contours.

### DIFF
--- a/Lib/trufont/windows/glyphWindow.py
+++ b/Lib/trufont/windows/glyphWindow.py
@@ -228,9 +228,10 @@ class GlyphWindow(BaseMainWindow):
     def selectAll(self):
         glyph = self.view.glyph()
         glyph.selected = True
-        if not len(glyph):
-            for component in glyph.components:
-                component.selected = True
+        for component in glyph.components:
+            component.selected = True
+        for anchor in glyph.anchors:
+            anchor.selected = True
 
     def deselect(self):
         glyph = self.view.glyph()


khaledhosny:
TruFont seems to be segfaulting on Travis, might be related to https://github.com/adrientetar/defconQt/commit/0b8c08bbbdf2eb2096ad080b9bf4cbe999618d62#commitcomment-18254911.

adrientetar:
> Currently it does not select anchors at all, and selects components only
when the glyph has no contours.

Yes, this is done on purpose. Neither of fontlab, glyphs, robofont do this afaict.

khaledhosny:
I don’t know, but it makes no sense to me. If I do select all, I really meant select all. Right now if I need to move the glyph around I have to go hunt for the components and the anchors, even worse I need to select each anchor and component individually since drag to select does not select them either. This is both non-intuitive and counter-productive IMO.

davelab6:
I agree with khaled

moyogo:
I asked a couple of designers here and they are used to how FontLab works. Sometimes they want to work on the outlines without touching the anchors. However they do find it annoying not being able to select anchors at anytime. TIL in Glyphs *Select All* selects the outlines, then *Select All* again selects all.

Other than that I totally agree with Khaled. I think the same goes for the selection rectangle.
Maybe this could be a configurable setting? Or alternatively Modifier key or Glyphs’ behaviour.

adrientetar:
> I don’t know, but it makes no sense to me. If I do select all, I really meant select all.

I do not disagree, it's just there seems to be a strong precedent of doing otherwise.

> TIL in Glyphs Select All selects the outlines, then Select All again selects all.

Thanks for asking around Denis, very useful. For now I implemented it like this in b4657d28a63ddc05f1e4f8fe5272ffe8db27efe9; we can revisit later if needed.

davelab6:
Glyphs's solution is a good compromise :) Thanks Adrien and Denis :)

PR #300, opened 2016-07-22
glyphWindow: glyphOffset() should skip glyphs in glyphOrder but not in font

moyogo:
In glyphWindow, Next Glyph or Previous Glyph may raise KeyError when a glyphOrder has glyphs not in a font.

### DIFF
--- a/Lib/trufont/windows/glyphWindow.py
+++ b/Lib/trufont/windows/glyphWindow.py
@@ -146,7 +146,13 @@ class GlyphWindow(BaseMainWindow):
         if not (glyphOrder and len(glyphOrder)):
             return
         index = glyphOrder.index(currentGlyph.name)
-        newIndex = (index + offset) % len(glyphOrder)
+        newIndex = index
+        sign = offset//abs(offset)
+        while True:
+            newIndex = newIndex + sign
+            newIndex = (newIndex) % len(glyphOrder)
+            if glyphOrder[newIndex] in font:
+                break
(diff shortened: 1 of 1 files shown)


adrientetar:
Hmm. I think what we should really do is making template glyphs for glyph that are in glyphOrder but not in font. I thought that was the current behavior already.

moyogo:
It seems that changed in https://github.com/trufont/trufont/commit/35b6257f0f39f18b668e45905b31d5a4f63762e6#diff-f2f869f598056034c344c35ea0834826L798

Having FontWindow._updateGlyphsFromGlyphOrder() do the following seems to do the trick:
```python
if glyphName in font:
    glyphs.append(font[glyphName])
else:
    glyphs.append(font.newStandardGlyph(glyphName, asTemplate=True))
```

But then, we might still want to skip over the template glyphs with Next Glyph and Previous Glyph.
For example, in https://github.com/daltonmaag/scope-one, we have template glyphs only to neatly arrange glyphs in rows of 32.

khaledhosny:
I don’t think next/previous glyph should normally skip template glyphs.

adrientetar:
3c33bc3bf8aa2113966b190a167dcc6d62783c25

> we have template glyphs only to neatly arrange glyphs in rows of 32.

Hum, that's not really the primary intent of template glyphs. However maybe this set should be configurable, if you want to open an issue for it.

PR #302, opened 2016-07-24
Use Unix line endings

khaledhosny:
This file is the only one in repository using DOS line endings, I think
we better stick to one for consistency.


anthrotype:
We could also check in a .gitattributes file with `*.py text eol=lf` to force Unix line endings.

PR #310, opened 2016-07-30
requirements.txt: add --find-links to download booleanOperations==0.4 from Github Releases page

anthrotype:
…


anthrotype:
Installing Trufont from source no longer requires a C++ compiler.

anthrotype:
well, at least for Windows and OS X. For Linux, I still haven't found the time to understand how to make a manylinux1 wheel.

Btw, have you noticed, PyQt 5.7 is out ;)

adrientetar:
Oh nice, I just did the upgrade.

Issue #385, opened 2017-03-09
Saving in the font features window should save the font


khaledhosny:
It is rather inconvenient to have so save the features, then switch the font window and save the font each time one modifies the features.

PR #389, opened 2017-03-09
Save the font when saving the features as well

khaledhosny:
Fixes https://github.com/trufont/trufont/issues/385.


madig:
What keeps this from being merged?

davelab6:
@khaledhosny would you merge this yourself now?

Issue #387, opened 2017-03-09
Glyph tab should always show the glyph name


khaledhosny:
Currently it seems if the glyph has a `unicode` value the character is shown, but I think the glyph name is still important and should always be shown. I suggest always showing the glyph name, then the character in brackets if present, e.g. “uni0640 (ـ)”, or something like that.

PR #391, opened 2017-03-09
Always show glyph name in the glyph tab title

khaledhosny:
If the glyph character (from the unicode value) is different than the
glyph name, show it in parenthesis as well.

Fixes https://github.com/trufont/trufont/issues/387.


madig:
What keeps this from being merged?

davelab6:
@khaledhosny I propose that you merge this :)

Issue #275, opened 2016-04-23
Pan in preview mode 


pathumego:
Hold spacebar to pan in Glyph view. 
This is requested in #176 #250 #258.

anthrotype:
Preferably, not just the space bar (like FF), but space bar *and* mouse click-drag, with the cursor toggling to "hand" tool (like Glyphs.app)

pathumego:
@anthrotype 👍

adrientetar:
You can click mouse wheel to pan. Spacebar to pan isn't yet implemented.

adrientetar:
baed42489c673e04e61303bd99c65b7ace2342e0

PR #397, opened 2017-03-24
[drawingTools] add a simple Hand tool to pan glyph canvas view

anthrotype:
This is my first experiment hacking with PyQt5 and the TruFont GUI, so please be kind ;)

It's somewhat related to https://github.com/trufont/trufont/issues/275

I thought we could make the space key toggle on/off this hand tool, so that we could support the filled-in preview combined with panning of the canvas on dragging the cursor (like Glyphs and Robofont do).


adrientetar:
baed42489c673e04e61303bd99c65b7ace2342e0

anthrotype:
without hands! 👍

Issue #337, opened 2016-08-27
Put simple install instructions in README


adrientetar:
The "simple way" should be something like:

```
pip3 install PyQt5
git clone https://github.com/trufont/trufont
cd trufont
pip3 install -r requirements.txt
python3 setup.py install
trufont
```

venv only needs be used in certain cases.

anthrotype:
`pip install trufont` FTW

adrientetar:
> pip install trufont FTW

Feel free to help with that ;)

anthrotype:
Looking at the list of Trufont's requirements, only defconQt and extractor are not yet on PyPI.
I'll do that in the following days.

anthrotype:
Arrgghh.. of course, "extractor" name is already taken on PyPI :(
https://pypi.python.org/pypi/Extractor/0.5

how shall I name the package? "ufo-extractor" maybe?

I wonder if it could be a good idea to merge extractor into ufo2ft... 
They are kind of complementary, one goes from fonttools objects to ufo, the other from ufo to fonttools objects.

anthrotype:
or even better, ufoLib itself...

adrientetar:
> how shall I name the package? "ufo-extractor" maybe?

Sounds good, yea. Thanks!

davelab6:
> I wonder if it could be a good idea to merge extractor into ufo2ft... or even better, ufoLib itself...

I like this idea, of having both ufo2ft and ft2ufo classes in ufoLib :)

anthrotype:
Let's think about that a bit more...

In the meantime, [defconQt](https://pypi.python.org/pypi/defconQt) and [extractor](https://pypi.python.org/pypi/ufo-extractor) are out.

I had to name extractor "ufo-extractor" to avoid the name clash.
So one would `pip install ufo-extractor`, but still `import extractor`.
Not ideal, but something to start with.

Now, on to TruFont itself.

anthrotype:
I pushed c8fd8dccc9176985e5c624e81fd117cff563f3d8

Next time you push a tag, it should trigger the `deploy` stage on Travis CI, which will publish the distribution packages to the Python index.

It also creates a Github Release from the pushed tag, if it doesn't already exist. No files are uploaded automatically to Github, at least for now.

Maybe we can make it so it runs pyqtdeploy to create the Mac and Linux app bundles (and same for Windows on Appveyor), but I don't know how to use pyqtdeploy.

adrientetar:
[...]

PR #407, opened 2017-03-29
Add TESTING.md

DeeDeeG:
These instructions are easy to follow, and should lower the barrier of entry for testing the GitHub/master branch of TruFont.

TODO: if anyone wants to make instructions for other distros or operating systems, that would make this more useful.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,7 +57,7 @@ If you want to test TruFont/master without making changes, see `TESTING <https:/
 Contributing
 ~~~~~~~~~~~~
 
-Here's a quick tutorial if you'd like to contribute to TruFont. 
+Here's a quick tutorial if you'd like to contribute to TruFont.
 
 1. Click the "Fork" button above, and clone the forked git repository
    to a new directory called ``trufont``:
--- a/README.rst
+++ b/README.rst
@@ -49,10 +49,15 @@ Getting started
 
 4. Run the app as ``trufont``.
 
+Testing
+~~~~~~~
+
+If you want to test TruFont/master without making changes, see `TESTING <https://github.com/trufont/trufont/blob/master/TESTING.md/>`__
+
 Contributing
 ~~~~~~~~~~~~
 
-Here's a quick tutorial if you'd like to contribute to TruFont.
+Here's a quick tutorial if you'd like to contribute to TruFont. 
 
 1. Click the "Fork" button above, and clone the forked git repository
    to a new directory called ``trufont``:
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,6 +1,8 @@
 ## Intro
 
-This is a step-by-step tutorial to get TruFont running on a LiveCD. This allows you to test TruFont easily, and is simpler / easier than the instructions in README.md. If all else fails, try these instructions.
+This is a step-by-step tutorial to get TruFont running on an Ubuntu live CD. This enables you to test the GitHub/master branch of TruFont easily, and is simpler than the instructions in README.md. 
+
+If all else fails, try these instructions.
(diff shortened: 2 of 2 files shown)


davelab6:
I don't think a new TESTING file is needed, but in running through the install instructions recently myself, I agree they need a little improvement. Here's my take:

https://github.com/trufont/trufont/compare/master...davelab6:davelab6-contributing?expand=1

DeeDeeG:
Sounds good. 

Some of the things I had in my tutorial are only needed from a Live CD, so I'll list what steps can be removed:

Don't need to recommend users install Git.

universe repository should be enabled after an install, only Live CD users would have to enable it.

Also, the  devs said not to run `sudo pip3` or it will break the overall system's Python configuration :neutral_face:. (That's probably why I had such a hard time installing in the first place.) Running from within a VENV, and using `pip3 install -e` which is the same as `pip3 install --editable` should make the permissions error go away.

DeeDeeG:
I think there's three different use-cases, and trying to cover all three maybe gets a bit unwieldy.

- Some users just want the latest "release" version, and they can use pip install trufont in a VENV. That downloads the latest release version from the Python package index.
- Some users want to hack on TruFont and be able to merge their changes. They can follow the current "contributing" section of the README.rst. Although I'd argue all the details about setting upstreams in git commandline and rebasing from upstream/master might not be needed.
- Some users don't care about merging any changes, but just want to test. They don't need to know how to set upstreams and rebase from master, as long as they can get TruFont to run.

davelab6:
On Mar 29, 2017 7:27 PM, "DeeDeeG" <notifications@github.com> wrote:

I think there's three different use-cases, and trying to cover all three
maybe gets a bit unwieldy.

   - Some users just want the latest "release" version, and they can use
   pip install trufont in a VENV. That downloads the latest release version
   from the Python package index.


They should use a binary package download, then.


   - Some users want to hack on TruFont and be able to merge their changes.
   They can follow the current "contributing" section of the README.rst.
   Although I'd argue all the details about setting upstreams in git
   commandline and rebasing from upstream/master might not be needed.

I think given the level of detail you want for below, it's fine to include
this.


   - Some users don't care about merging any changes, but just want to
   test. They don't need to know how to set upstreams and rebase from master,
   as long as they can get TruFont to run.


That seems like step 1 to the above step 2, though?

DeeDeeG:
These are a lot of subjective decisions that I would like to leave to the maintainers.

In terms of binary package download vs `pip install trufont`, both are equivalent, in that they let you run the release version, but @anthrotype expressed a preference for ultimately having users do `pip install trufont` where possible, in this thread: #337 

And for myself, on Linux the `pip install` way ties in better to the command-line. You can just type `trufont` and it will run. Although with the binary download you don't need a venv. I leave this to the maintainers, it's not really my place to decide.

Speaking of #337 ... Maybe it can just be a simple "install from source" heading on README.rst, like in your version, but with the commands @adrientetar listed in comment 1. My TESTING.md assumes a completely fresh install of Ubuntu, which is not keeping with the rest of README.rst. But the commands he listed are more similar to the level of detail used elsewhere in this repo.

DeeDeeG:
Whoops... I feel like facepalming right now, because all this effort is pretty much for nothing.

All one has to do to get this running on Ubuntu is clone from github, and run `pip3 install --editable .` from inside trufont dir.

In other words, following the the README.rst "contributing" steps 1 and 6, although I think step 6 should be changed to say pip3 install, rather than pip install.

My Python was really messed up on the other machine so everything kept failing, but it was my fault for screwing up Python on my other machine, probably.

PR #451, opened 2017-05-29
Open new fonts in current font window if it contains an empty font.

### LINKED ISSUE #433: Open file in current window if it's empty
When starting Trufont, I usually open a UFO next, which is then opened in a new window. I then must close the old one. I know about "Load last file on start", but that gets me unwanted behavior when handling multiple UFOs.

So how about loading files into the current window when the current font is empty, like on startup?

madig:
Before, starting Trufont and opening some font yielded two windows: One
with an empty font and the other you opened. Change this to open the
selected font into the window with the empty font if it hasn't been
modified, saving you one Alt+Tab and close-window.

Closes #433.

### DIFF
--- a/Lib/trufont/objects/application.py
+++ b/Lib/trufont/objects/application.py
@@ -461,12 +461,21 @@ class Application(QApplication):
                     widget.raise_()
                     return
         font = TFont()
+        currentFont = self.currentFont()
         try:
             font.extract(path)
         except Exception as e:
             errorReports.showCriticalException(e)
             return
-        window = FontWindow(font)
+        # Open new font binary in current font window if it contains an
+        # unmodified empty font (e.g. after startup).
+        if currentFont.path is None and currentFont.binaryPath is None \
+           and currentFont.dirty is False:
+            window = self.currentFontWindow()
+            window.setFont_(font)
+            window.setWindowTitle(window.fontTitle())
+        else:
+            window = FontWindow(font)
         window.show()
         self.setCurrentFile(font.binaryPath)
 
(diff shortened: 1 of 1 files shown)


adrientetar:
Like #389 this is Product Manager stuff but doing this breaks the UX consistency that one window is one document, that makes the program more confusing. The real fix would probably be adding a welcome screen to the program where one can open a file/recent file etc.

davelab6:
I disagree; it's confusing that the empty document that is unwanted stays
around

On May 29, 2017 5:51 PM, "Adrien Tétar" <notifications@github.com> wrote:

Like #389 <https://github.com/trufont/trufont/pull/389> this is Product
Manager stuff but doing this breaks the UX consistency that one window is
one document, that makes the program more confusing. The real fix would
probably be adding a welcome screen to the program where one can open a
file/recent file etc.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/trufont/trufont/pull/451#issuecomment-304735608>, or mute
the thread

madig:
I just tried LibreOffice Writer and GNOME's GEdit. They actually behave like what this patch is trying to do. As long as the starting (empty) document is unchanged, loading a file will replace the empty document with it. Wow, I don't remember being aware of this behavior, it just felt natural.

Glyphs has it somewhat easier, being a macOS app. Starting it just presents the global menu, the user has to open a new font or load one by himself.

davelab6:
Adrien, do you know of any apps that do what you suggest, other than
FontForge?

madig:
Well, what's the state here? I'm still for merging this.

adrientetar:
2f5730895757a5475be24fdf0f96d23f8856bf26

adrientetar:
This doesn't refresh opened widgets with the new font.

davelab6:
> This doesn't refresh opened widgets with the new font.

I don't understand what you mean

moyogo:
>>This doesn't refresh opened widgets with the new font.

>I don't understand what you mean

Trufont on HEAD doesn’t start with a new empty font and if a new empty font is open, opening an existing UFO replaces the new font.

madig:
Hm. As widgets and such should contain no modified state (otherwise, the state would be "dirty" and the font would be loaded into a new window), notifiying the widgets should be enough?

Issue #423, opened 2017-05-01
Adding, deleting, sorting, reverting glyphs or renaming anchors randomly crashes Trufont on Linux


madig:
1. Open any font
2. Start deleting glyphs one after the other until Trufont segfaults.

Backtrace: [backtrace.txt](https://github.com/trufont/trufont/files/968386/backtrace.txt)

madig:
It seems there is a common cause for the crashes that I'm seeing. The same backtrace also comes up when sorting and adding glyphs and renaming anchors. So #420 is probably related.

madig:
I don't get it.

Still trying to find the root cause for all the crashes I get on up-to-date Fedora and Ubuntu. I shortened addGlyphs (https://github.com/trufont/trufont/blob/master/Lib/trufont/windows/fontWindow.py#L915) to:
```
    def addGlyphs(self):
        glyphs = self.glyphCellView.glyphs()
        newGlyphNames, params, ok = AddGlyphsDialog.getNewGlyphNames(
            self, glyphs)
        if ok:
            sortFont = params.pop("sortFont")
            for name in newGlyphNames:
                glyph = self._font.get(name, **params)
        #         if glyph is not None:
        #             glyphs.append(glyph)
        #     self.glyphCellView.setGlyphs(glyphs)
        #     if sortFont:
        #         # TODO: when the user add chars from a glyphSet and no others,
        #         # should we try to sort according to that glyphSet?
        #         # The above would probably warrant some rearchitecturing.
        #         # kick-in the sort mechanism
        #         self._font.sortDescriptor = self._font.sortDescriptor
```
... and glyphs still get added normally and Trufont still crashes, although less often it seems? 

`self._font.get()` indirectly invokes `TLayer.get()` (https://github.com/trufont/trufont/blob/master/Lib/trufont/objects/defcon.py#L188) which *creates* glyphs and returns them? Is this on purpose?

madig:
Just tested this on Windows 10 1703 with Python 3.6.1 32-bit and 32-bit Qt 5.8.2 installed via pip, no crash in several rounds of adding glyphs and sorting them back and forth.

On Ubuntu 17.04 64-bit, I get the same crashes as on Fedora 25 64-bit. Sigh.

madig:
The backtrace I get now mentions something new, `QWindow::activeChanged()`:
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
[...]

Issue #333, opened 2016-08-15
Can’t open any menus


khaledhosny:
On Gnome 3 the menus close immediately whenever I open them (it is closed so quickly most of time that it appears like nothing is happening).

adrientetar:
Can you bisect? I'll try but I have Cinnamon in the office and I need to install all the deps.

khaledhosny:
Bisected to 65a6a0a4316216d4774e9dae8864cd6978f9546d.

adrientetar:
Interesting, so if you use the current code but changing to focusChanged it works?

khaledhosny:
Yes, after reverting that commit menus are back to normal.

adrientetar:
Ok I can repro.

adrientetar:
0d6862ff29effc2e70df22223ef40c17a37c8f24

adrientetar:
There's also a segfault on exit, ha. time to look at that too.

madig:
@khaledhosny: Can you describe your issue again? The workaround for this bug leads to crashes: #423. I can't quite reproduce what you wrote...

khaledhosny:
I think the issue summary is as clear as it gets, not sure what else needs explanation. That being said, I can’t reproduce the issue anymore after reverting 0d6862ff29effc2e70df22223ef40c17a37c8f24.

madig:
Note to self: Should this issue come up again, try saving the window hash instead of a reference.

PR #463, opened 2017-06-06
Revert "application: fix menu switch"

madig:
This reverts commit 0d6862ff29effc2e70df22223ef40c17a37c8f24.

The original bug can no longer be reproduced (see #333); additionally,
the workaround caused #423. Saving the last window with focus caused
random crashes on Linux when certain dialogs were closed, an issue with
object lifetimes may be involved.


adrientetar:
Have you tried saving the window hash instead of a reference?

madig:
I haven't, but it's an idea if the original bug should crop up again.

PR #465, opened 2017-06-07
Fix Extension Builder window

khaledhosny:
This dialog is extremely unusable, whatever I do will result in assert or error somewhere! With these fixes I can finally build an extension and register a menu item (though I’m yet to get that menu item to do anything, but that is a different issue).


adrientetar:
5ea489c138e80ff5a9e32b4f7b5f02c0bc6fcf50

Still want version to throw if the input is invalid.

PR #489, opened 2017-09-09
layoutEngine: Guard harder against feaLib failures

khaledhosny:
Apparently feaLib will happily compile feature text using non-existent
glyphs, but compiling the table will obviously fail, so the compilation
step needs to be inside the try statement as well.

Without this change editing any glyphs in Reem Kufi source UFO will fail
as there are missing glyphs referenced by the feature code (they are
added at build time by the build script).

### DIFF
--- a/Lib/trufont/objects/layoutEngine.py
+++ b/Lib/trufont/objects/layoutEngine.py
@@ -24,9 +24,9 @@ def _layoutEngineOTLTablesRepresentationFactory(layoutEngine):
     if font.features.text:
         otf = TTFont()
         otf.setGlyphOrder(glyphOrder)
-        # compile with feaLib + markWriter. kerning is handled separately
+        # compile with feaLib + markWriter/kernWriter
         try:
-            compiler = FeatureCompiler(font, otf, kernWriterClass=None)
+            compiler = FeatureCompiler(font, otf)
             compiler.postProcess = lambda: None
             compiler.compile()
             for name in ("GDEF", "GSUB", "GPOS"):
@@ -127,7 +127,6 @@ class LayoutEngine(BaseObject):
         funcs.set_nominal_glyph_func(_get_nominal_glyph, None, None)
         funcs.set_glyph_h_advance_func(_get_glyph_h_advance, None, None)
         # TODO: vertical advance
-        # TODO: kerning funcs from UFO kerning?
         funcs.set_glyph_name_func(_get_glyph_name_func, None, None)
         font.set_funcs(funcs, self, None)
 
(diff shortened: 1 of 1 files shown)


Lib/trufont/objects/layoutEngine.py
```diff
@@ -29,17 +29,16 @@ def _layoutEngineOTLTablesRepresentationFactory(layoutEngine):
...
             for name in ("GDEF", "GSUB", "GPOS"):
                 if name in otf:
                     table = otf[name].compile(otf)
                     value = hb.Blob.create_for_array(
                         table, HB.MEMORY_MODE_READONLY)
                     ret[name] = value
+        except:
+            # TODO: handle this in the UI
+            import traceback
+            print(traceback.format_exc(5))
```
adrientetar: Should we do something there? maybe ret = dict() to drop already collected tables? maybe we don't want

khaledhosny: I guess so.

BTW, there is also the case when the feature text changes, do you drop the existing tables in this case? If not the we should.

adrientetar: See beginSelfFeaturesObservation()

adrientetar: I think what's not implemented yet is UFO kerning and the text tool should call the layout engine again when spacing or kerning is changed.

khaledhosny: For kerning I thought we could use a `get_glyph_h_kerning` callback, tried that now and it does it work and now I remember that HarfBuzz won’t use this calback if the font has a GPOS table. So I guess we need to process the kerning with feaLib after all. Hopefully this will not be a big issue, generating GSUB and GPOS without kerning seems to be pretty fast already.

Lib/trufont/objects/layoutEngine.py
```diff
@@ -24,9 +24,9 @@ def _layoutEngineOTLTablesRepresentationFactory(layoutEngine):
     if font.features.text:
         otf = TTFont()
         otf.setGlyphOrder(glyphOrder)
-        # compile with feaLib + markWriter. kerning is handled separately
+        # compile with feaLib
```
adrientetar: well now it's feaLib + kernWriter/markWriter

adrientetar: writers are in ufo2ft not feaLib

adrientetar:
How about adding kerning values manually i.e. post processing harfbuzz results?

khaledhosny:
That might work, but I’m not sure it is worth the trouble. May be consider it if compiling kerning in GPOS turned out to be considerably slow.

adrientetar:
Yea that's what I'm thinking.

PR #496, opened 2017-10-06
Fix "TypeError: newGlyph() missing 1 required positional argument"

anthrotype:
After this commit 3df77f720b7f2bcfe01eaf4112b36b8bcb3f8462, I am seeing this error every time I try to import an OTF or TTF, which uses extractor.

The reason is we can't just assign `Font.newGlyph`, which is a class attribute, and thus an "unbound" method, to an instance as self.newGlyph.

In order to dynamically "bind" a method to an instance, we can use `types.MethodType`.

https://docs.python.org/3/library/types.html

Also, the original newGlyph method is restored with a try/finally block, in case anything goes wrong during the extraction.

### DIFF
--- a/Lib/trufont/objects/defcon.py
+++ b/Lib/trufont/objects/defcon.py
@@ -9,6 +9,7 @@ from trufont.objects.undoManager import UndoManager
 from ufo2ft import compileOTF, compileTTF
 import fontTools
 import math
+import types
 
 _shaper = True
 try:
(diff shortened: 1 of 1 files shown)


adrientetar:
Thanks. yea I know about method binding but it worked without it when I wrote the code, I don't know why it didn't occur.

anthrotype:
Are you sure it wasn't happening on you? I'm using Linux with python 3.6.3. It would be strange if it behaves differently elsewhere...

adrientetar:
I just tried and I do see the error, so I guess I didn't do it properly last time..

Issue #382, opened 2017-03-08
Drag-and-drop SVG files


brawer:
It would be nice if one could drag-and-drop SVG files (like [these](https://github.com/googlei18n/noto-emoji/tree/master/svg)) into TruFont. When dropping an SVG into the glyph editor, it would add the path geometry from the dragged SVG file; when dropping an SVG into the font view, it would add the glyph. See https://github.com/googlei18n/noto-emoji/issues/61#issuecomment-284862828 by @DeeDeeG.

DeeDeeG:
Wow, I can't believe this is suddenly supported! Thanks @santhoshtr and everyone who reviewed/merged.

colmoneill:
Hi, I'm trying to use this feature, I have v0.6.5 running here, but I can't seem to find any way of pasting in svg to TruFont in any way. Is this feature enabled yet? Am I missing something a step in the process? Is this feature only available with development versions of TruFont?
I really hope I can work this out, I'd love to use TruFont.
Thanks and all the best,

santhoshtr:
Works fne for me. Here is a video showing it in action: https://imgur.com/0KNK3dy 
I am at commit:  bb0548cf899146269d3ef00cc1df47e3bf86f587

Do you see any errors while you drag and drop?

PR #499, opened 2017-10-14
Support pasting SVG from clipboard

santhoshtr:
Paste an svg in clipboard to trufont glyph editor. Note that it is not the filename in clipboard, but the SVG copied from an svg editor.

This helps me for my workflow. I draw and edit my glyphs in Inkscape. Now I can just select the drawing in Inkscape, copy and paste in Trufont.

I referred https://github.com/fonttools/fonttools/blob/master/Snippets/svg2glif.py

Addressing #382 (Drag and drop svgs) should be more easy after this patch.

Known issue: Pasted svg appears vertically flipped. svg2glif produce glif output with vertical flip.

### DIFF
--- a/Lib/trufont/windows/fontWindow.py
+++ b/Lib/trufont/windows/fontWindow.py
@@ -6,6 +6,8 @@ from PyQt5.QtWidgets import (
 from defconQt.controls.glyphCellView import GlyphCellView
 from defconQt.windows.baseWindows import BaseWindow
 from fontTools.feaLib.error import FeatureLibError
+from fontTools.misc.py23 import SimpleNamespace
+from fontTools.svgLib import SVGPath
 from trufont.controls.exportDialog import ExportDialog
 from trufont.controls.fileMessageBoxes import CloseMessageBox, ReloadMessageBox
 from trufont.controls.fontDialogs import AddGlyphsDialog, SortDialog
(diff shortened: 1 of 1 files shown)


Lib/trufont/windows/fontWindow.py
```diff
@@ -1048,6 +1069,29 @@ def _proceedWithDeletion(self, erase=False):
...
+        glyph = SimpleNamespace(width=width, height=height, unicodes=unicodes)
+        outline = SVGPath.fromstring(svg, transform=transform)
+
+        # writeGlyphToString takes a callable (usually a glyph's drawPoints
+        # method) that accepts a PointPen, however SVGPath currently only has
+        # a draw method that accepts a segment pen. We need to wrap the call
+        # with a converter pen.
+        def drawPoints(pointPen):
+            pen = SegmentToPointPen(pointPen)
+            outline.draw(pen)
```
adrientetar: You don't need to call the adapter pen manually, you can just do glyph.getPen() instead of getPointPen().

Lib/trufont/windows/fontWindow.py
```diff
@@ -1048,6 +1069,29 @@ def _proceedWithDeletion(self, erase=False):
...
+
+        return writeGlyphToString(name,
+            glyphObject=glyph,
+            drawPointsFunc=drawPoints,
+            formatVersion=version)
```
adrientetar: I'm not sure why you write to a glyph then read from string right after, also why creating another glyph object here when the caller already has glyph/otherGlyph. I think even otherGlyph can be removed in this instance as the SVGPath object can draw itself into a pen.

So you can do svg.draw(glyph.getPen()). Anyway, I'll do the merge. Thanks!

adrientetar:
78109b2a004cadaeb4a247c3330d69bdb2311c71

PR #501, opened 2017-10-14
Export SVG to the clipboard as well

khaledhosny:
This way one can copy glyphs from TruFont and paste them into Inkscape. We might want to support other vector formats for popular vector graphics editors.

### DIFF
--- a/Lib/trufont/windows/fontWindow.py
+++ b/Lib/trufont/windows/fontWindow.py
@@ -7,6 +7,7 @@ from defconQt.controls.glyphCellView import GlyphCellView
 from defconQt.windows.baseWindows import BaseWindow
 from fontTools.feaLib.error import FeatureLibError
 from fontTools.svgLib import SVGPath
+from fontTools.pens.svgPathPen import SVGPathPen
 from trufont.controls.exportDialog import ExportDialog
 from trufont.controls.fileMessageBoxes import CloseMessageBox, ReloadMessageBox
 from trufont.controls.fontDialogs import AddGlyphsDialog, SortDialog
@@ -666,6 +667,7 @@ class FontWindow(BaseWindow):
                 glyph.clear()
 
     def copy(self):
+        font = self._font
         widget = self.stackWidget.currentWidget()
         clipboard = QApplication.clipboard()
         mimeData = QMimeData()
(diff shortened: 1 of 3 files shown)


khaledhosny:
Depends on https://github.com/fonttools/fonttools/pull/1073.

anthrotype:
@khaledhosny I hope you don't mind, I've renamed your ``SVGPen`` to its original ufo2svg name, ``SVGPathPen``:
https://github.com/fonttools/fonttools/commit/76e49d6ff5dfbaba6d9f1f577f8a4c5298014b92

I think the latter name describes it better, just SVG sounds too broad.

khaledhosny:
I see, I though the “Path” part is about the 2 different types of pens (point and segment? I never understood the difference). Thanks for correcting it.

Lib/trufont/windows/fontWindow.py
```diff
@@ -678,13 +684,40 @@ def copy(self):
             packGlyphs = (
                 glyphs[index] for index in sorted(
                     self.glyphCellView.selection()))
+
+        fontHeight = font.info.ascender - font.info.descender
+        glyphSet = {n: font[n] for n in font.glyphOrder}
```
adrientetar: You can just pass the font instead of copying it.

Lib/trufont/windows/fontWindow.py
```diff
@@ -30,6 +30,11 @@
 import os
 import pickle
 
+try:
+    from fontTools.pens.svgPen import SVGPathPen
+except ImportError:
+    SVGPathPen = None
```
adrientetar: Let's wait for a new pip and remove conditional code.

Lib/trufont/windows/fontWindow.py
```diff
@@ -678,13 +684,40 @@ def copy(self):
...
+        svgGlyphs = []
         pickled = []
+        x = 0
         for glyph in packGlyphs:
             pickled.append(glyph.serialize(
                 blacklist=("name", "unicodes")
             ))
+
+            if SVGPathPen is not None:
+                pen = SVGPathPen(glyphSet)
```
adrientetar: Can we not recreate the pen? There should be a clear method or sth.

Lib/trufont/windows/fontWindow.py
```diff
@@ -678,13 +684,40 @@ def copy(self):
...
+                width = glyph.width or glyph.bounds[2] - glyph.bounds[0]
+                g = '<g transform="matrix(1,0,0,-1,%f,0)"><path d="%s"/></g>' \
+                    % (x, pen.getCommands())
+                svgGlyphs.append(g)
+                x += width
+
         mimeData.setData("application/x-trufont-glyph-data",
                          pickle.dumps(pickled))
+
+        if svgGlyphs:
```
adrientetar: This will be non None if we always use svg pen.

Lib/trufont/windows/fontWindow.py
```diff
@@ -678,13 +684,40 @@ def copy(self):
             packGlyphs = (
                 glyphs[index] for index in sorted(
                     self.glyphCellView.selection()))
+
+        fontHeight = font.info.ascender - font.info.descender
```
adrientetar: font.info.upm?

adrientetar: unitsPerEm

Lib/trufont/windows/fontWindow.py
```diff
@@ -678,13 +684,40 @@ def copy(self):
...
+        x = 0
         for glyph in packGlyphs:
             pickled.append(glyph.serialize(
                 blacklist=("name", "unicodes")
             ))
+
+            if SVGPathPen is not None:
+                pen = SVGPathPen(glyphSet)
+                glyph.draw(pen)
+                width = glyph.width or glyph.bounds[2] - glyph.bounds[0]
```
adrientetar: Combining marks might intentionally have zero width, but that doesn't matter.

adrientetar: maybe*

Lib/trufont/windows/fontWindow.py
khaledhosny: That should work.

Lib/trufont/windows/fontWindow.py
khaledhosny: I keep forgetting that 😅

Lib/trufont/windows/fontWindow.py
khaledhosny: Does it matter, most of time we are copying one glyph, and even if one is copying multiple glyphs it is unlikely to be an issue.

Lib/trufont/windows/fontWindow.py
khaledhosny: That was specifically to handle them. When copying multiple glyphs, I was putting them next to each other but marks would then overlap. I changed this and I now put the glyphs in a grid of five columns using the upm for all glyphs, it looks nicer this way.

adrientetar:
Okay I think you can 3.17 the requirements.txt, that should fix the build.

khaledhosny:
> Okay I think you can 3.17 the requirements.txt, that should fix the build.

Done and `setup.py` too since we won’t work with older versions anyway.

pathumego:
🎉

PR #507, opened 2017-10-24
Allow editing anchor coordinates on double click

khaledhosny:
Adjusting the exact position of an anchor is often needed, and without
this patch I find myself opening the glif files in a text editor to edit
the anchors.

### DIFF
--- a/Lib/trufont/controls/glyphDialogs.py
+++ b/Lib/trufont/controls/glyphDialogs.py
@@ -1,4 +1,5 @@
-from PyQt5.QtCore import QEvent, Qt
+from PyQt5.QtCore import QEvent, QLocale, Qt
+from PyQt5.QtGui import QDoubleValidator
 from PyQt5.QtWidgets import (
     QDialog, QDialogButtonBox, QGridLayout, QLabel, QLineEdit, QListWidget,
     QRadioButton)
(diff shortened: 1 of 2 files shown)


Lib/trufont/controls/glyphDialogs.py
```diff
@@ -162,17 +163,26 @@ def getLayerAndAction(cls, parent, currentGlyph):
...
         self.setWindowModality(Qt.WindowModal)
-        self.setWindowTitle(self.tr("Rename…"))
+        self.setWindowTitle(self.tr("Edit…"))
 
         nameLabel = QLabel(self.tr("Name:"), self)
         self.nameEdit = QLineEdit(self)
         self.nameEdit.setFocus(Qt.OtherFocusReason)
 
+        validator = QDoubleValidator(self)
+        validator.setLocale(QLocale.c())
```
adrientetar: Is this to input numbers in the user locale's format? If that's needed it should probably be done consistently throughout because I don't think that's done anywhere else at the moment.

khaledhosny: I just wanted to restrict the input to numbers, if you know a better way I can use it.

adrientetar: QDoubleValidator forces the input to be a number yes, but the locale thing isn't called anywhere else we use QDoubleValidator. The current locale should be set by default already. Why do you call setLocale?

khaledhosny: We don’t currently use `QDoubleValidator` anywhere else, and I wasn’t sure the rest of the code can handle locale-specific float formatting (but if you use French locale yourself then the code probably handles it 😄)

khaledhosny: Oh, I remember now. I use an Arabic locale, and without forcing C locale here the input box won’t accept entering `.` as decimal point and would only accept the Arabic decimal separator which actually is not accessible from Arabic keyboard layout.

adrientetar:
I'll probably do this differently in the future, but for now I'll merge it in.

PR #530, opened 2018-01-05
Support drag and dropping .glif files to glyph canvas

santhoshtr:
Helpful for using glif files created from svg for example.

If the glyph has components, and they should pre-exist in UFO.

### DIFF
--- a/Lib/trufont/controls/glyphCanvasView.py
+++ b/Lib/trufont/controls/glyphCanvasView.py
@@ -12,6 +12,7 @@ from trufont.objects import settings
 from trufont.objects.layoutManager import LayoutManager
 from trufont.tools import drawing, errorReports
 from trufont.tools.uiMethods import UIGlyphGuidelines
+from ufoLib.glifLib import readGlyphFromString
 import os
 
 GlyphViewMinSizeForGrid = 10000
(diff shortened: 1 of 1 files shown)


adrientetar:
> One caveat is if the glyph has components, this has no effect.

Why not?

santhoshtr:
If the glyph that is being dragged has components like below, there is no points to draw. That is what I meant.
```
  <outline>
    <component base="glif1"/>
    <component base="glif2" xOffset="2049" yOffset="-382"/>
  </outline>
```

adrientetar:
drawPoints draws components as well, only anchors and guidelines are left-out.

davelab6:
@adrientetar do you want further changes to this, or are you rejecting it?

santhoshtr:
I changed

> One caveat is if the glyph has components, this has no effect.

to
> If the glyph has components, and they should pre-exist in UFO.

I hope this clarifies the confusion in PR description. Do you need any code improvements from me?

Lib/trufont/controls/glyphCanvasView.py
```diff
@@ -305,7 +306,7 @@ def dragEnterEvent(self, event):
             if url.isLocalFile():
                 path = url.toLocalFile()
                 ext = os.path.splitext(path)[1][1:]
-                if ext.lower() in QImageReader.supportedImageFormats():
+                if ext.lower() in QImageReader.supportedImageFormats() + ['glif']:
```
adrientetar: line too ling

Issue #535, opened 2018-01-22
Can we update the README to use git and not pip?


EwonRael:
Right now cloning from git seems like the way to get TruFont, so can the instructions in the README reflect this instead of using pip?

Small thing I know

eliheuer:
I always install in editable mode, even when I'm using TruFont to work on UFOs. This is what I usually do:

```
python3 -m venv ENV_DIR
source ENV_DIR/bin/activate
cd ENV_DIR
git clone git@github.com:trufont/trufont.git
cd trufont
pip install --upgrade -r requirements.txt
pip install --editable .
trufont
```
Is that what you are suggesting @EwonRael?

davelab6:
I think it's a good idea

eliheuer:
@EwonRael I made a new branch in my fork with updated install instructions: https://github.com/eliheuer/trufont/tree/issue-535

Let me know if this looks good to you and I will make a pull request. 

Also, I made a YouTube video demonstrating the quick install chain method, which I use **all the time**: https://youtu.be/sLvm5185_ak

madig:
Installing from PyPI is now the way to go, unless you want to hack on TruFont.

PR #539, opened 2018-01-25
Update README to fix Issue 535

eliheuer:
This is an attempt to fix issue #535. The installation method using PyPI is not working for me on Arch and Ubuntu. @EwonRael is also experiencing problems.

I think that editable mode is the best feature of TruFont and could be made more discoverable.


anthrotype:
I don't have issues if you like to make the editable installation more prominent in the README.
But the fact that `pip3 install trufont` doesn't Just Work is just our laziness at tagging regular releases and uploading distribution packages to PyPI.
This should be made as simple as pushing a git tag: the setup.py version should be computed automatically from the latter (using setuptools_scm, e.g. see [here](https://github.com/fonttools/ttfautohint-py/commit/f3f3a826e09008688db9cf9e82db8e45b3642e0c)), and Travis CI should take care of deploying both the wheel and sdist to PyPI only for tagged commits. We do this elsewhere, I just haven't found the time to this here (and am not sure Adrien is interested?).

Also, I imagine that not all users wish to hack the font editor's source code (but all are welcome), and for them I would still like to recommend downloading the stable release, perhaps without any venv but only a global `--user` install option (of course, provided that we do tag new releases more often).

eliheuer:
Ok thanks, I'm looking for ways to contribute, so If you or Adrien need help keeping things updated let me know what I can do to help. 

I'll see if I can setup the setuptools_scm/Travis deploy in a fork using https://testpypi.python.org/pypi

anthrotype:
the travis deploy is tricky because you need to have push access to generate the token. I'll try to work on that if I find the time. In the meantime, maybe you could cook up a PR for the setuptools_scm thinghy?
thanks for your contributions!

README.rst
```diff
@@ -4,53 +4,114 @@ TruFont
...
-   -  Linux: It's usually packaged with the OS.
+   -  Linux: It's usually packaged with the OS,
+      run ``python3 --version`` to see what you have installed. 
 
-2. Set up a new Python **virtual environment**. Although this is not
-   required, it's *highly recommended*, since TruFont has lots of
+2. Install **Git 2.1** (or later):
+
+   -  OS X: Install using `Homebrew <http://brew.sh/>`__:
+      ``brew install git``
```
davelab6: I believe git is a dependency of Homebrew and installed by the Homebrew installation shell script

README.rst
```diff
@@ -4,53 +4,114 @@ TruFont
...
       ``brew install python3``
    -  Windows: Download installer from 
       `python.org/downloads <https://www.python.org/downloads/>`__
-   -  Linux: It's usually packaged with the OS.
+   -  Linux: It's usually packaged with the OS,
+      run ``python3 --version`` to see what you have installed. 
 
-2. Set up a new Python **virtual environment**. Although this is not
-   required, it's *highly recommended*, since TruFont has lots of
+2. Install **Git 2.1** (or later):
```
davelab6: Perhaps, `Install **Git** (v2.1+):`?

Also perhaps add a line to say how to check if you have it already?

README.rst
```diff
@@ -4,53 +4,114 @@ TruFont
...
-written with Python3, ufoLib, defcon and PyQt5.
+written with Python3, ufoLib, defcon, and PyQt5.
+
+.. image:: Lib/trufont/resources/screenshot.png
 
 Getting started
-~~~~~~~~~~~~~~~
+===============
+
+Quick installation instructions
```
davelab6: I would insert the `pip install trufont` instruction here, with an explanation of its limitation (depending on the developer team making release packages.)

README.rst
```diff
@@ -4,53 +4,114 @@ TruFont
...
+     python3 -m venv trufont-venv
+     source trufont-venv/bin/activate
+     cd trufont-venv
+     git clone https://github.com/trufont/trufont.git
+     cd trufont
+     pip install --upgrade -r requirements.txt
+     pip install --editable .
+     trufont
+
+Detailed Installation Instructions
```
davelab6: Maybe rename this Contributor Installation Instructions and include the above 'all on 1 line' and 'all at once' commands, then go on to explain them?

davelab6:
@eliheuer thanks for this!

I think its important to get this merged soon as I'd like to tell people about the project at upcoming type design conferences, starting with TypoDay, the national Indian type design event

khaledhosny:
I’d rather we cut a new release than asking everyone to install from Git (this isn’t FontForge after all). @adrientetar what is needed to make a new release?

davelab6:
> I’d rather we cut a new release than asking everyone to install from Git

To be clear, I think `pip3 install trufont` should be at the top as the most prominent way to install; and 'contributor' instructions could even be moved to a `FAQ.md` or `DEVELOPER.md` file :)

> (this isn’t FontForge after all)

I believe this is (or means to be) a self-extensible editor, and therefore showing people how to install complete sources and edit them is an important user journey. These installation instructions are just the first step in that journey :)

pathumego:
Either way it would be nice to have a new release out :)

README.rst
davelab6: Hmm seems not any more! this is still needed I think

eliheuer:
This is no longer relevant for the WX branch/version so I'm closing it.

PR #558, opened 2018-04-23
Drop macOS Travis CI, use Python 3.6

madig:


anthrotype:
old macos have an outdated SSL library which causes issues with latest PyPI. Curl'ing the get-pip.py script and installing/upgrading pip from there should work.

Also `brew upgrade python` _should_ just work.

anthrotype:
if you're gonna use the Mac Python.org, please install 3.6.5

.travis.yml
```diff
@@ -13,15 +14,19 @@ matrix:
...
   - echo $PATH
   - which python3
+  - python3 --version
   - pip3 --version
   - pip3 list
   # install requirements
-  - pip3 install --upgrade pip setuptools wheel
+  - python3 -m venv venv
+  - . venv/bin/activate
+  - curl https://bootstrap.pypa.io/get-pip.py | python3
```
anthrotype: pipe the command into `python`, as you're already inside the venv

madig:
I have no problem with jumping to 3.6 for all CI work. @adrientetar?

Also, what exactly do we need the macOS CI for, anyway?

.travis.yml
```diff
@@ -13,15 +14,19 @@ matrix:
...
+  - pip3 install --upgrade setuptools wheel
```
anthrotype: looks like you don't get `pip3` when running get-pip.py on linux: https://travis-ci.org/trufont/trufont/jobs/370693421#L502

just call it `pip` (as you're in the venv) or `python3 -m pip`

anthrotype:
> I have no problem with jumping to 3.6 for all CI work

we were already using 3.6 with homebrew python

> what exactly do we need the macOS CI for

it's more like a wish

madig:
So... Maybe we can simplify the process by dropping macOS for now?

Do we want fully equipped binary wheels for all platforms?

The Python packaging story for user applications is a bit underwhelming. I
suppose we need more scaffolding for providing binaries for Windows and
macOS?

anthrotype:
> we can simplify the process by dropping macOS for now?

i'm ok to drop it for now. It's creating more issues than it solves.

> Do we want fully equipped binary wheels for all platforms?

there isn't anything "binary" in TruFont itself, the package is pure python.

pathumego:
Thanks guys! :)

madig:
What I mean is ready-to-execute binary packages for the release page, so a dmg/exe/flatpak that may contain Python 3.6.

Issue #598, opened 2018-10-17
WX: Opening Ubuntu-B fails


madig:
Ubuntu-Bold ([test.zip](https://github.com/trufont/trufont/files/2489176/test.zip)), turned into a .tfont with the following code:
```Python
import tfont.converters
tf = tfont.converters.UFOConverter().open("Ubuntu-B.ufo")
tfont.converters.TFontConverter(indent=2).save(tf, "/tmp/test.tfont")
```
On opening the file in Trufont-wx, I get repeated exceptions:
```
Traceback (most recent call last):
  File "/home/nikolaus/Entwicklung/trufont/src/trufont/controls/glyphCellView.py", line 329, in OnPaint
    drawing.drawGlyphFigure(ctx, glyph, cellWidth, cellHeight, color)
  File "/home/nikolaus/Entwicklung/trufont/src/trufont/util/drawing.py", line 160, in drawGlyphFigure
    ctx.FillPath(layer.closedComponentsGraphicsPath, wx.WINDING_RULE)
  File "/home/nikolaus/Entwicklung/tfont/src/tfont/objects/layer.py", line 160, in closedComponentsGraphicsPath
    return self.closedComponentsGraphicsPathFactory()
  File "/home/nikolaus/Entwicklung/trufont/src/trufont/objects/factories.py", line 47, in layerClosedComponentsGraphicsPathFactory
    graphicsPath.AddPath(component.closedGraphicsPath)
  File "/home/nikolaus/Entwicklung/tfont/src/tfont/objects/component.py", line 68, in closedGraphicsPath
    return self.closedGraphicsPathFactory()
  File "/home/nikolaus/Entwicklung/trufont/src/trufont/objects/factories.py", line 11, in componentClosedGraphicsPathFactory
    graphicsPath.AddPath(component.layer.closedGraphicsPath)
  File "/home/nikolaus/Entwicklung/tfont/src/tfont/objects/layer.py", line 167, in closedGraphicsPath
    self.closedGraphicsPathFactory()
  File "/home/nikolaus/Entwicklung/trufont/src/trufont/objects/factories.py", line 55, in layerClosedGraphicsPathFactory
    graphicsPath.AddPath(path.graphicsPath)
  File "/home/nikolaus/Entwicklung/tfont/src/tfont/objects/path.py", line 61, in graphicsPath
    graphicsPath = self._graphicsPath = self.graphicsPathFactory()
[...]

PR #599, opened 2018-10-17
WX: Handle quadratic curves

madig:
Fixes https://github.com/trufont/trufont/issues/598.

http://unifiedfontobject.org/versions/ufo3/glyphs/glif/#contour says a contour only containing off-curves must be treated as quadratic. Still need to implement that.

Also: more resilience to invalid curves like 3 off-curves and a `curve`? Ideally, this is caught and the UI displays the glyph as containing invalid contours, allowing the user to go in and fix it...

### DIFF
--- a/src/trufont/objects/factories.py
+++ b/src/trufont/objects/factories.py
@@ -97,6 +97,17 @@ def pathGraphicsPathFactory(path):
             if point.type == "curve":
                 graphicsPath.AddCurveToPoint(*stack)
                 stack = []
+            elif point.type == "qcurve":
+                if len(stack) == 4:
+                    graphicsPath.AddQuadCurveToPoint(*stack)
+                else:
+                    for offset in range(0, len(stack) - 4, 2):
+                        cx1, cy1, cx2, cy2 = stack[offset : offset + 4]
+                        graphicsPath.AddQuadCurveToPoint(
+                            cx1, cy1, (cx1 + cx2) / 2, (cy1 + cy2) / 2
+                        )
+                    graphicsPath.AddQuadCurveToPoint(*stack[-4:])
+                stack = []
     if not open_:
         graphicsPath.CloseSubpath()
     return graphicsPath


madig:
Not sure my handling of all-off-curve contours is clean?

madig:
Also, I'd love to have a unit test for this :thinking:

adrientetar:
Thanks, I have exams this week but I'll look at it in the we

madig:
With https://github.com/trufont/tfont/pull/8 I can open the glyphs and they look pretty crappy. Mixture of quadratic and cubic curves in e.g. `B`.

madig:
Oops, because I don't handle quadratic off-curves between two non-smooth points, they get misdetected as cubics.

src/trufont/objects/factories.py
```diff
@@ -96,7 +96,36 @@ def pathGraphicsPathFactory(path):
             stack.append(point.y)
             if point.type == "curve":
                 graphicsPath.AddCurveToPoint(*stack)
-                stack = []
+                stack.clear()
```
adrientetar: stack = [] does the same, but does not need a method lookup.

src/trufont/objects/factories.py
```diff
@@ -96,7 +96,36 @@ def pathGraphicsPathFactory(path):
...
+            #     path.
+            elif point.type == "qcurve" or point is points[-1]:
+                if len(stack) == 1:
+                    graphicsPath.AddQuadCurveToPoint(*stack)
+                    stack.clear()
+                else:
+                    stack_offcurves = trufont.util.misc.grouper(stack, 2)
+                    for (cx1, cy1), (cx2, cy2) in trufont.util.misc.pairwise(
+                        stack_offcurves
+                    ):
```
adrientetar: - This assumes number of points is even, right?
- That's a lot of intermediary functions for a simple iteration, we can instead do something like:

```python
offset = 0
size = len(stack)
while 4 + offset < size:
  points = stack[offset:4+offset]
  ...
  offset += 2
```

Can be rewritten in terms of for loop as e.g.:

```python
for offset in range(0, len(stack) - 4, 2):
  points = stack[offset:4+offset]
```

adrientetar:
- How common are "all off curves" paths? Which tools produce this and which file formats and editors support it?
- This will need some changes in tfont first,
  * add support for qcurve in segment-specific ops of tfont, in path.py. search for pts_len
  * in tfont slice.py, the slicing code depends on the presence of segments, and segments are formed by the presence of an on curve point.  
just an example of code that "all off curves" paths will break in tfont and trufont
- I'd rather we keep "all off curves" out at least for now

madig:
> How common are "all off curves" paths? Which tools produce this and which file formats and editors support it?

Dunno, but the UFO spec talks about it. Can do a `if all(offcurve): explode()` for now? Speaking of explosions, I'd like TruFont to handle exceptions while rendering more gracefully than now, where it basically corrupts the UI and probably the entire application. Todo.

madig:
I just realized that the `len(stack) == 1` line should actually be `== 4`. D'oh.

madig:
This actually renders glyphs correctly now. Thanks for the algorithm!

Now I need to look at handles being drawn without a line if we're looking at qcurves. Anything else?

I can't seem to find a way to properly test the paths though :(

madig:
Heh.
```
  File "/home/nikolaus/Entwicklung/tfont/src/tfont/objects/path.py", line 281, in bounds
    "cannot compute bounds for %r segment" % self.type)
NotImplementedError: cannot compute bounds for 'qcurve' segment
```

src/trufont/objects/factories.py
```diff
@@ -76,13 +76,16 @@ def pathGraphicsPathFactory(path):
     points = path._points
     if not points:
         return graphicsPath
+    if all(point.type is None for point in points):
```
adrientetar: This is a hot code path. Why make a linear time assertion for a possibility that we don't know is possible?

src/trufont/objects/factories.py
```diff
@@ -76,13 +76,16 @@ def pathGraphicsPathFactory(path):
...
+        # https://github.com/trufont/trufont/pull/599#issuecomment-441391756
+        raise ValueError("All-off-curve paths are not currently supported.")
+
     start = points[0]
     open_ = skip = start.type == "move"
     if open_:
         graphicsPath.MoveToPoint(start.x, start.y)
     else:
         start = points[-1]
-        assert start.type is not None
```
adrientetar: Can you bring this back, a path is a suite of segments and a segment ends with an oncurve.

src/trufont/objects/factories.py
```diff
@@ -97,6 +100,36 @@ def pathGraphicsPathFactory(path):
             if point.type == "curve":
                 graphicsPath.AddCurveToPoint(*stack)
                 stack = []
+            elif point.type == "qcurve" or (not open_ and point is points[-1]):
```
adrientetar: what's (not open_ and point is points[-1]) ? what should the last point be assumed to be qcurve?

src/trufont/objects/factories.py
```diff
@@ -97,6 +100,36 @@ def pathGraphicsPathFactory(path):
...
+                assert len(stack) % 2 == 0, "Stack must have even number of points."
```
adrientetar: Comment is basically the same thing as the code before, so unneeded. If the assert fails, the LOC will be included in the traceback either way.

src/trufont/objects/factories.py
```diff
@@ -97,6 +100,36 @@ def pathGraphicsPathFactory(path):
...
+                #     looking at the qcurve: stack == [cx, cy, qx, qy], ready to
+                #     AddQuadCurveToPoint.
+                #  2. We have more than one off-curve on the stack and are looking at
+                #     the last element, the qcurve. Here, TrueType's implied on-curve
+                #     principle applies that requires that an implied on-curve point
+                #     is inserted halfway between all off-curve points. The last
+                #     off-curve is then used as the control-point of the qcurve.
+                #  3. The stack is full of off-curves and we have reached the last
+                #     point. An all off-curve path is to be handled like a quadratic
+                #     path.
```
adrientetar: This is far too long and I think unneeded

src/trufont/util/drawing.py
```diff
@@ -440,9 +440,14 @@ def drawLayerPoints(
             # others
             prev = points[-1]
             breakHandle = path.open
+            quadraticPath = any(point.type == "qcurve" for point in points)
```
adrientetar: ugh. this whole code is written to read the path only once. why is this needed?

src/trufont/util/drawing.py
```diff
@@ -440,9 +440,14 @@ def drawLayerPoints(
...
             for point in points:
                 isOff = point.type is None
-                if not breakHandle and isOff != (prev.type is None):
+                if (
+                    not breakHandle
+                    and isOff != (prev.type is None)
+                    and not quadraticPath
+                ):
```
adrientetar: do we even need to change the code that works for curve?

adrientetar:
> Dunno, but the UFO spec talks about it.

If you don't know, then don't bother. There's far enough work with things that actually happen, and every new LOC in the codebase is extra maintenance and cognitive burden for the next person working with that code. And they'll be misled into thinking that it's something that's actually expected.

UFO spec tends to go and "support every possibility" but it's more harmful than not. It weakens invariants about the data and makes code more complicated than need be. If someone needs a feature, they'll open a ticket and explain why they need it. Good software is always driven by needs and requests &ndash; feature set and code as short as can be.

Likewise about assertions, only assert something if it's a case that would subtly corrupt the output and be hard to debug or even realize it's wrong. And it oughta be a simple thing. e.g. I assert stack is empty when doing a line otherwise the state machine would accept offcurve followed by oncurve which is an instant GTFO

src/trufont/util/drawing.py
adrientetar: ok I see the commit is named "hide handles". why hide? can you add a picture of what you want to do with handles?

src/trufont/objects/factories.py
madig: Oops. Can't remember why I deleted this.

src/trufont/util/drawing.py
madig: It's a cosmetic thing mostly, I find drawn handles on qcurve off-curves confusing because they suggest curve semantics. Leave in instead? Do something else?

src/trufont/objects/factories.py
madig: It's part of the all-off-curve handling. Remove completely?

src/trufont/util/drawing.py
adrientetar: it's not curve semantics, it's smooth point semantics

adrientetar: I would leave as-is personally

src/trufont/objects/factories.py
adrientetar: for the "all off curves" thing probably.

src/trufont/objects/factories.py
madig: Hm. Guess it can be turned into a `assert not stack` after the point loop?

src/trufont/objects/factories.py
adrientetar: You only need to assert this in the else below

madig:
The last assert doesn't seem to be needed actually as the other asserts catch all-off-curves already.

src/trufont/objects/factories.py
```diff
@@ -97,6 +97,20 @@ def pathGraphicsPathFactory(path):
             if point.type == "curve":
                 graphicsPath.AddCurveToPoint(*stack)
                 stack = []
+            elif point.type == "qcurve":
+                if len(stack) == 4:
+                    graphicsPath.AddQuadCurveToPoint(*stack)
+                else:
+                    # Group off-curves into quads to construct implied on-curve points.
+                    assert len(stack) % 2 == 0
```
adrientetar: this can be removed, elements are always added in pairs so this is always true

src/trufont/objects/factories.py
```diff
@@ -97,6 +97,20 @@ def pathGraphicsPathFactory(path):
             if point.type == "curve":
                 graphicsPath.AddCurveToPoint(*stack)
                 stack = []
+            elif point.type == "qcurve":
+                if len(stack) == 4:
+                    graphicsPath.AddQuadCurveToPoint(*stack)
+                else:
+                    # Group off-curves into quads to construct implied on-curve points.
```
adrientetar: "Group off-curves into quads" it's more like splitting than grouping, but I would say "make simple quads by constructing implied oc" because that's what's happening. large segment broken down to simple ones

src/trufont/objects/factories.py
```diff
@@ -97,6 +97,20 @@ def pathGraphicsPathFactory(path):
...
+                    assert len(stack) % 2 == 0
+                    for offset in range(0, len(stack) - 4, 2):
+                        cx1, cy1, cx2, cy2 = stack[offset : offset + 4]
+                        graphicsPath.AddQuadCurveToPoint(
+                            cx1, cy1, (cx1 + cx2) / 2, (cy1 + cy2) / 2
+                        )
+                    # Use last two points as actual control and on-curve point.
```
adrientetar: can be removed, it's obvious from the code

src/trufont/objects/factories.py
madig: I don't understand that sentence 😬

adrientetar: By constructing implied oc (`(cx1 + cx2) / 2, (cy1 + cy2) / 2`) we can make simple quads (a segment made of just one off followed by on which is the canonical form)

adrientetar:
Btw, why do you need this? you wouldn't draw with quads?

src/trufont/objects/factories.py
```diff
@@ -97,6 +97,19 @@ def pathGraphicsPathFactory(path):
             if point.type == "curve":
                 graphicsPath.AddCurveToPoint(*stack)
                 stack = []
+            elif point.type == "qcurve":
+                if len(stack) == 4:
+                    graphicsPath.AddQuadCurveToPoint(*stack)
+                else:
+                    # Iterate over off-curve points in quads (current, next) to
```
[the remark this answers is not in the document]
madig: Then away the comment goes.

madig:
No, but I occasionally handle sources with converted curves and inspect TTFs.

src/trufont/objects/factories.py
madig: Oh, ok. I thought by "quad" you meant the four elements in the stack slice.

adrientetar:
(This still needs supporting changes to tfont, to be clear)

adrientetar:
??
Why did you merge this without the supporting changes?

madig:
Uh, well, it allows me to more easily test stuff as there's one less branch to juggle. Revert?

adrientetar:
No need at this time but I mean let's not merge half-baked things unless we need to.

madig:
Ok, sorry.

adrientetar:
np, and thanks for the PR and for addressing all the review comments 🙂

madig:
I just found a real-world example for an all-off-curves glyph: https://github.com/typesupply/defcon/blob/master/Lib/defcon/test/testdata/TestFont.ufo/glyphs.public.background/A_.glif

adrientetar:
> I just found a real-world example

No, you found defcon test suite.

moyogo:
https://github.com/TypeNetwork/Amstelvar was drawn with quadratic curves.

adrientetar:
Are there tools that have a quadratic pen tool?

jenskutilek:
If you want to work with TrueType hinting at some point, you should take extreme care

1. not to change the point indexes of imported outlines
2. to support everything that the glyf table spec allows, including all-offcurve contours.

madig:
I just found an all-off-curve qcurve in the wild 😲 

DejaVu Sans ExtraLight, in the bowls of `six` and `nine`.

PR #612, opened 2018-11-21
Support opening non-tfont files

madig:
Convert UFOs on the fly to tfont objects.

### DIFF
--- a/src/trufont/objects/application.py
+++ b/src/trufont/objects/application.py
@@ -1,7 +1,8 @@
+import os
 from trufont import __version__
 from trufont.windows.fontWindow import FontWindow, prepareNewFont
 from typing import List, Optional
-from tfont.converters import TFontConverter
+from tfont.converters import TFontConverter, UFOConverter
 from tfont.objects import Font
 from weakref import WeakSet
 import wx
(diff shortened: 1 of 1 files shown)


src/trufont/objects/application.py
```diff
@@ -113,18 +113,35 @@ def openFont(self, path=None) -> Optional[Font]:
...
                     return
-        font = TFontConverter().open(path)
+
+        # Load and convert the file and append it to the file history.
+        if path.endswith(".tfont"):
+            font = TFontConverter().open(path)
+        elif path.endswith((".ufo", ".ufoz")):
+            font = UFOConverter().open(path)
+        elif path.endswith("metainfo.plist"):
+            import os.path
```
adrientetar: Here you can probably just import os at the top of the file.

madig:
I don't like how wxDialog.GetPath does not append the selected file extension. Will re-dig into this once I can save UFOs...

src/trufont/objects/application.py
```diff
@@ -1,7 +1,8 @@
+import os.path
```
adrientetar: like I said, import os is enough

src/trufont/windows/fontWindow.py
```diff
@@ -247,7 +247,8 @@ def _newTab(self, glyph=None):
     def save(self, path=None):
         if path is None:
             path = self._path
-        if path is None:
+        # XXX: Expand to handle more formats once converters can properly round-trip.
+        if path is None or not path.endswith(".tfont"):
```
adrientetar: Can you just not set a path into the window if you're not gonna save to that path? Because that's really the point of _path. Besides, only native format should be a Ctrl+S target. Conversion to another format is potentially expensive/lossy so should not be done if not called for.

src/trufont/windows/fontWindow.py
```diff
@@ -600,6 +597,8 @@ def OnSaveAs(self, event):
             if dialog.ShowModal() == wx.ID_CANCEL:
                 return
             path = dialog.GetPath()
+            if not path.endswith(".tfont"):
+                path += ".tfont"
```
adrientetar: This is unfortunate, because say if you're saving to foo the dialog has not checked that foo.tfont doesn't already exist, if which case it'll be overwritten silently. I don't think I've needed this line on windows, but if it's needed elsewise...

madig: Hrm. Maybe I need to ask around about this...

src/trufont/windows/fontWindow.py
madig: So, do you want non-tfont import to be done specifically from an "import" menu entry with an "export" entry to go along with it?

madig:
I decided to separate this out into it's own `importFont()` method. Let's see. Fighting with recently-opened-click segfaults...

src/trufont/objects/application.py
```diff
@@ -108,24 +109,47 @@ def newFont(self) -> Font:
         FontWindow(None, font).Show()
         return font
 
-    def openFont(self, path=None) -> Optional[Font]:
+    def openFont(self, path: Optional[str] = None) -> Font:
```
adrientetar: The return type is Optional[Font] if you look at the method body

madig: Yes, caught that after mypy gave me a weird look.

madig:
Treating open and import separately leads to a slight if-else increase.

adrientetar:
I don't think we need to have separate open and import,

madig:
What about opening binary OT fonts?

adrientetar:
Thanks.

PR #614, opened 2018-11-27
[WIP] Various thoughts about implementing undo/redo for TruFont

justvanrossum:
Probably best to use this PR as a public discussion before merging.

The document is mostly a brainstorm session about things that could and should, and does not yet contain a very concrete guideline about what to do.

At the same time, BlackFoundry has been experimenting with an undo system in their fork (referenced in my doc). It would be great to learn from their experience as well.

### DIFF
--- a/docs/undo.md
+++ b/docs/undo.md
@@ -2,57 +2,106 @@
 
 ## Introduction to Undo Systems
 
-An undo/redo stack is often implemented with as an “undo manager” class. In the case of an outline editor, a lot of data will simply be copied, as it’s not always possible to simply “perform the reverse operation”. Transformations are a good example of that. But also point behavior: when a handle is implicitly being moved because it has a “smooth” connection to another, the opposite movement may result in a different position of the “implied” movement due to rounding. So in general, we need to efficiently copy the complete previous state of an outline to be able to do undo reliably.
+An undo/redo stack is often implemented with as an “undo manager” class.
+In the case of an outline editor, a lot of data will simply be copied, as
+it’s not always possible to simply “perform the reverse operation”.
+Transformations are a good example of that. But also point behavior: when
+a handle is implicitly being moved because it has a “smooth” connection
+to another, the opposite movement may result in a different position of
+the “implied” movement due to rounding. So in general, we need to
+efficiently copy the complete previous state of an outline to be able to
+do undo reliably.
 
-If diffing between versions can be done with text-representations of objects, the github.com/google/diff-match-patch library may be of interest; Python’s difflib is not really an option as it only does diffing, not patching.
+If diffing between versions can be done with text-representations of
+objects, the github.com/google/diff-match-patch library may be of
+interest; Python’s difflib is not really an option as it only does
+diffing, not patching.
 
-Perhaps much of the code can be implemented as a separate library, for example called “tundo”.
+Perhaps much of the code can be implemented as a separate library, for
+example called “tundo”.
 
 ## Considerations about resource usage
 
-Should all undo data be kept in RAM (possibly limiting the number of undo levels to be kept) or should undo data be flushed to disk, to minimize RAM usage?
+Should all undo data be kept in RAM (possibly limiting the number of undo
+levels to be kept) or should undo data be flushed to disk, to minimize
+RAM usage?
 
 ## Factoring
 
-From the perspective of the UI and object code, the undo API should be simple. Objects may have to provide efficient way to serialize themselves (either as binary data or as text), but the undo manager will take care of storing data efficiently. The undo API will have no notion about how efficiently or inefficiently the undo data is stored. Whether data is stored in memory or on disk will be completely opaque to the undo client. This will allow us to build a working system first, and then measure and optimize for performance and resource usage.
+From the perspective of the UI and object code, the undo API should be
+simple. Objects may have to provide efficient way to serialize themselves
+(either as binary data or as text), but the undo manager will take care
+of storing data efficiently. The undo API will have no notion about how
+efficiently or inefficiently the undo data is stored. Whether data is
+stored in memory or on disk will be completely opaque to the undo client.
+This will allow us to build a working system first, and then measure and
+optimize for performance and resource usage.
 
 ## Undo vs Selection
 
-Undo is not only about the model data, but also selection data: if you do a step back with undo, the user also expects the selection to be the same as before. This notion is important because it implies undo data is more than just serialization of glyph data.
+Undo is not only about the model data, but also selection data: if you do
+a step back with undo, the user also expects the selection to be the same
+as before. This notion is important because it implies undo data is more
+than just serialization of glyph data.
 
-Perhaps the tfont library should support a serialization format that is specialized for undo, in that it contains selection info. I need to figure out how this needs to be factored. Perhaps the “converter” concept can be used.
+Perhaps the tfont library should support a serialization format that is
+specialized for undo, in that it contains selection info. I need to
+figure out how this needs to be factored. Perhaps the “converter” concept
+can be used.
 
 ## Undo vs Redo
 
-Each UndoManager should maintain an undo stack and a redo stack. When editing data, before the edit is applied, the current version of the data should be added to the undo stack. Performing undo is also an edit operation, but it’s special in that the current state of the data is now added to the redo stack. The design of the undo manager should make implementing this behavior easy.
+Each UndoManager should maintain an undo stack and a redo stack. When
+editing data, before the edit is applied, the current version of the data
+should be added to the undo stack. Performing undo is also an edit
+operation, but it’s special in that the current state of the data is now
+added to the redo stack. The design of the undo manager should make
+implementing this behavior easy.
 
 ## Undo vs Notifications
 
-Depending on how notifications are done in TruFont, some leverage may be gained from pairing undo with the notification mechanism. A more general notification system may allow this more easily. (Needs more investigation: see the trufont.util.tracker module, and how it’s used in layer.paths.)
+Depending on how notifications are done in TruFont, some leverage may be
+gained from pairing undo with the notification mechanism. A more general
+notification system may allow this more easily. (Needs more
+investigation: see the trufont.util.tracker module, and how it’s used in
+layer.paths.)
 
 ## Undo Manager API concepts
 
 Flow of events when editing:
 
-- An edit action is about to happen (for example the selection will be dragged)
-- The undo manager is told the name of the operation (In this case “drag”, so the Undo menu can say “Undo drag” instead of just “Undo”)
-- The undo manager is supplied with the current state of the data, which it will push onto the internal undo stack
+- An edit action is about to happen (for example the selection will be
+  dragged)
+- The undo manager is told the name of the operation (In this case
+  “drag”, so the Undo menu can say “Undo drag” instead of just “Undo”)
+- The undo manager is supplied with the current state of the data, which
+  it will push onto the internal undo stack
 - The edit action is handled (the user drags the selection)
-- The undo manager is told the edit action is done (not sure this is even needed)
+- The undo manager is told the edit action is done (not sure this is even
+  needed)
 - The internal redo stack is cleared
 
 Flow of events when the user does Undo:
 
-- The undo manager is informed that will will perform an undo step. This allows to record subsequent edits into the redo stack: it will enter redo-mode.
-- The undo manager is supplied with the current state of the data, which it will push onto the internal redo stack. If the previous action was another undo or a redo action, existing serialized data could be reused.
-- The data of the state to be restored is popped off the undo stack, and applied to the live data
-- The undo manager is informed that we’re done, allowing it to exit redo-mode
+- The undo manager is informed that will will perform an undo step. This
+  allows to record subsequent edits into the redo stack: it will enter
+  redo-mode.
+- The undo manager is supplied with the current state of the data, which
+  it will push onto the internal redo stack. If the previous action was
+  another undo or a redo action, existing serialized data could be
+  reused.
+- The data of the state to be restored is popped off the undo stack, and
+  applied to the live data
+- The undo manager is informed that we’re done, allowing it to exit
+  redo-mode
 
 Flow of events when the user does Redo:
 
 - The undo manager is informed that will will perform an redo step.
-- The undo manager is supplied with the current state of the data, which it will push onto the internal undo stack
-- The data of the state to be restored is popped off the redo stack, and applied to the live data
+- The undo manager is supplied with the current state of the data, which
+  it will push onto the internal undo stack
+- The data of the state to be restored is popped off the redo stack, and
+  applied to the live data
 - The undo manager is informed that we’re done.
 
 The undo manager should have methods to query its state:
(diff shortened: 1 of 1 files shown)


docs/undo.md
```diff
@@ -0,0 +1,280 @@
...
+investigation: see the trufont.util.tracker module, and how it’s used in
+layer.paths.)
+
+## Undo Manager API concepts
+
+Flow of events when editing:
+
+- An edit action is about to happen (for example the selection will be
+  dragged)
+- The undo manager is told the name of the operation (In this case
```
madig: And possibly more meta data, e.g. the concerned layer, so the user knows which drag he's undoing. Maybe this info can even be used for e.g. a visual indicator effect in the UI.

adrientetar:
Thanks for the comment. Agree that incremental undo is doable & desirable (I started to experiment with it in my C# app). One thing is the need to bubble-up the change, as there are various caches (e.g. there is a cached GUI path object in the model's Path - that should be cleared when there's a point change. Similarly when a change occurs the Glyph should be timestamped for change, etc. so the change goes up the tree all the way to the glyph, which stores the undo actions).

I had started playing with a change record, e.g. if you do point.x = 25 it'll create a change record such as this:

```c#
interface IChange {
  Time { get; } = new Unix.CurrentTime();

  Undo() {
    Debug.Assert(_obj.value == _newValue);
    _obj.value = _newValue;
    _obj.Parent?.ApplyChange(this, addToStack=false);
  }

  Redo() {
    ...
  }
}

public struct PointXChange : IChange
```

I guess the difference with your suggestion is you wanna implement the undo/redo logic in the objects themselves and not in the change class ?

I was thinking of storing a reference to the object such that I don't have to go down the tree (e.g., if you apply a change from the Glyph's undo stack you would have to know which path was this change made in and which point? if you store a reference to the changed point you can go up the tree again instead of down).

For collections, something like INotifyCollectionChanged's [NotifyCollectionChangedEventArgs](https://docs.microsoft.com/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs?view=netframework-4.7.2) is all that's needed to stash changes made to a collection (it has Action - move/replace/delete etc. -, NewItems, NewStartingIndex, OldItems, OldStartingIndex which can model any change to a collection).

>  When the user does an action through the UI, it creates a delta.

Note that you can also perform actions as part of executing a script, that's why I thought of a model where point.x = 25 creates a change record internally and sends it up to the undo manager (so no need to explicit care for undo on the user's part). But I understand your suggestion is to calculate the list of all changes to be made and then apply them?

It'd be cool if you can add a simple code sketch, given the the tree of objects (Glyph -> Path -> Point changes get stored inside glyph). As I see it there are two kinds of changes, value change and collection change.

I'll watch the talk you linked, thansk.

docs/undo.md
```diff
@@ -0,0 +1,280 @@
+# What should an undo system be like for TruFont?
+
+## Introduction to Undo Systems
+
+An undo/redo stack is often implemented with as an “undo manager” class.
+In the case of an outline editor, a lot of data will simply be copied, as
+it’s not always possible to simply “perform the reverse operation”.
+Transformations are a good example of that. But also point behavior: when
+a handle is implicitly being moved because it has a “smooth” connection
```
adrientetar: There's no implicit movement. Every movement is performed by an algorithm. So every movement can be recorded. YOu don't even have to round in your data class, you can round in your display pipeline.

YvesDup:
first of all, the following implementation affects only to the layer of a glyph. Other glyph modifications will be done later with the same model.

This work was done in three parts:

1 - A new class UndoRedoMgr as a classical undoredo manager (URM) with two stacks:  a undo stack and a redo stack

        A new action is pushed on the undo stack (on the top). An action is a tuple of 4 items: an operation, a function to run when undoes, a function to run when redoes and extra datas.

            function to run is stored as a functools.partial object
            extra datas are are the two snapshots of the layer - before and after operation. These datas are used in the functools.partial objects and to write unit tests.

        When undoes, action is pop from the undo stack and pushes to the redo stack only if undo function execution is correct (Else a DialogBox message will be display and a cleaning of stacks will done). The execution function will be controlled by a context manager (Not implemented yet)

        When redoes action, action moves from redo stack to undo stack after a succeed execution of redo function

2 - There is a creation of a new class UndoRedoGlyph which is

        a subclass of tfont.glyph

        and adding a UndoRedoMgr object as attribut

    When opening a font, each glyph is loaded in UndoRedoGlyph. The UndoRedoMgr will create only when the glyph is really open in TruFont

3 - Each target action in the code (as transformation, alignment, move, select path …) is added to the URM with a decorated function « layer_decorate_undoredo  » that runs the following instructions :

        Get a snapshot of the layer before operation

        Run the operation (this is the decorated function)

        Get an another snapshot of the modified layer

        Create an action and add it to the UMR

    Sometimes action can be placed in two different functions (as a move selection operation with mouse). We can use two joined decorators, « prepare_layer_decorate_undoredo  » to initiate the mechanism, and « perform_layer_decorate_undoredo » to finalize it

    So if a part of code to be « undoredoable » is inside a big function, I isolate these instructions in a new target function. And add it the appropriate decorator.

    A traditional approach is also possible, inserting code at the good place around the « action » instructions. This code is the same as describe below, in the first decorator « layer_decorate_undoredo  »

    The main advantage of the decorator is that the code is never modified and do not add new bug. But a decorator is hard to develop

The UndoRedoMgr class and all associated decorators are located in the objects.undoredomgr.py file.

The UndoRedoGlyph class is located in the objects.undoredoglyph.py file

Samples of implementation could be find in controls.propertiesView.py (a single decorator) and drawning.pentools.py (the twin decorators)

We talked with Jeremy and Samuel about the idea to include an UMR directly in tfont but we do not think that the perfect place, because there is no undo/redo function in font system.

At this time, my point of view is that work is a solid approach that must be finalize as soon as possible. But we have to write more unit tests on this part to improve the solution.

Decorator solution give us ability to enable or disable quickly the UMR, so we can code first a « new TruFont functions » and connect it later with the UMR.

YvesDup:
Possible optimizations:

First is about an option regarding the URM decorators of TruFont . All of them are distinct functions but could be rewrite in a single decorator as a class (with magic __call__ method). This option has to harden the solution. It could develop later when all functionalities of TruFont are done.

Second is about to reduce the structure of extra datas in the UMR. For each action we stored  two snaphots of datas (layer before and layer after). But It is possible to stored only one snashot and using that of previous action - or next action -  when undoes or redoes. 
So all the snapshots could be in the dedicated sequence (list or set, I do not know yet) of UMR

docs/undo.md
```diff
@@ -0,0 +1,280 @@
...
+
+This approach should be compared to a more state-based approach, where one
+freezes the state of an object, which can then be used to reinstate that
+version of the object later. Serialization fits a model of version control
+better, but may be less efficient.
+
+## TruFont Qt Undo Manager
+
+The previous generation of TruFont (based on Qt) has an undo manager
+written by Lasse Fister:
```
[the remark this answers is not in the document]
justvanrossum: Fixed.

docs/undo.md
justvanrossum: True, I'll reformulate.

But what about transformations? If one rotates an entire glyph, would one record the changes to all points? What I was trying to say there is that recording the inverse transformation is problematic, as it will cause rounding errors in floating point numbers.

justvanrossum:
@YvesDup thank you for your detailed message about your current work. I think indeed it needs to be avoided that *two* snapshots are taken upon every operation.

Right now, the only other comment I have is that I would not use the word "Redo" in class and module names. Undo implies Redo, so "UndoManager" should be good enough. Likewise, I'd call the module undoManager.py, to stick with the naming conventions used in the project.

justvanrossum:
@raphlinus thank you so much for sharing your ideas. I need to get my head around how this would work in a practical implementation, but I think I understand the main ideas. I'm considering making a toy implementation to study your proposal in more detail.

docs/undo.md
justvanrossum: I think this would only be needed if one can edit multiple layers at the same time. If that's not the case, one can only undo whatever happened in the active layer, so it should be clear what will happen.

raphlinus:
@justvanrossum Certainly. And apologies for not following up to Adrien's questions earlier; I've been slammed with other things (it'll hopefully be clear soon why), and writing code samples would take a fair amount of time to prepare. I think writing a toy implementation is an excellent idea, and I'd be more than happy to take a look at it, or answer specific questions - this week is shaping up to be better than last to focus.

adrientetar:
> A JSON path can be actiual jsonpath,

Okay reading this again I now understood that jsonpath [is a thing](http://jsonpath.com/) (kinda like xml queries?), there's also [jsonpatch](http://jsonpatch.com/) which is a standard syntax for JSON deltas.

> I'm considering making a toy implementation to study your proposal in more detail.

I'll be trying it also.

docs/undo.md
adrientetar: I see, I thought you meant rounding as in the font editor rounds values to the grid. My bad.

I agree that doing floating point arithmetic in undo isn't great - that's why I rather record oldValue and newValue.

> If one rotates an entire glyph, would one record the changes to all points?

I'd do that. Also keeps the set of possible operations simpler

raphlinus:
On closer look, I think [JSON Pointer](https://tools.ietf.org/html/rfc6901) is probably more appropriate than jsonpath, we don't need a rich query language, just a way to identify particular points in the tree. Also, I think jsonpatch could be useful if we want to serialize the undo records to disk (for example, being able to undo after quitting and restarting) but that serialization is not essential.

justvanrossum:
> On closer look, I think JSON Pointer is probably more appropriate than jsonpath [...]

That seems the same (or similar) to the jsonpatch paths that Adrien referred to.

However, for a Python implementation I think it makes more sense to represent a path as a tuple of path elements, saving the need for string parsing.

I like the overall design of jsonpatch, and I think I'll use that as a model for my experiments.

adrientetar:
[Here's a demo](https://gist.github.com/adrientetar/c687caa8c02d604c942d24d4eb9c8746) with Path/Point and Path has an undo stack and a move(dx) method.

It works the only thing is now instead of writing

```py
for point in path.points:
    point.x += 2
```

One writes

```py
changes = []
for ix, point in enumerate(path.points):
    changes.append(Change(
        "replace", f"/points/{ix}", {"x": point.x + 2}
    ))
path.apply(changes)
```

There must be a way to simplify syntax with some side effects e.g. annotating the first example

```py
# path.changes = [] (initial state)
for ix, point in enumerate(path.points)
    # add '/ix' to incoming changes
    point.x = 2
    # push change {'x': 2}

# apply changes
```

YvesDup:
@just: regarding the UndoManager, I did the job. So the project is
    up to date now
    About your first request, we have to fix few bugs in the actual
    version of TruFont before to do these optimizations. 
    So, it will be done
    Merry Christmas
    Cheers
    
    Yves 
    
    Le 17/12/2018 à 12:49, Just van Rossum
      a écrit :
    
    
      
      @YvesDup
        thank you for your detailed message about your current work. I
        think indeed it needs to be avoided that two snapshots
        are taken upon every operation.
      Right now, the only other comment I have is that I would not
        use the word "Redo" in class and module names. Undo implies
        Redo, so "UndoManager" should be good enough. Likewise, I'd call
        the module undoManager.py, to stick with the naming conventions
        used in the project.
      —
        You are receiving this because you were mentioned.
        Reply to this email directly, view it on GitHub, or mute the thread.
      {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/trufont/trufont","title":"trufont/trufont","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/trufont/trufont"}},"updates":{"snippets":[{"icon":"PERSON","message":"@justvanrossum in #614: @YvesDup thank you for your detailed message about your current work. I think indeed it needs to be avoided that *two* snapshots are taken upon every operation.\r\n\r\nRight now, the only other comment I have is that I would not use the word \"Redo\" in class and module names. Undo implies Redo, so \"UndoManager\" should be good enough. Likewise, I'd call the module undoManager.py, to stick with the naming conventions used in the project."}],"action":{"name":"View Pull Request","url":"https://github.com/trufont/trufont/pull/614#issuecomment-447818117"}}}
      [
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/trufont/trufont/pull/614#issuecomment-447818117",
"url": "https://github.com/trufont/trufont/pull/614#issuecomment-447818117",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

justvanrossum:
I continued with my undo manager, and it is now available in its own repo: https://github.com/justvanrossum/jundo

I really like described the system @raphlinus described.

In my code I focussed less on the "single-source-of-truth" concept (although my code does not preclude taking that route), and more on "how can this be exposed to a Python programmer in a convenient form". Also: "how can undo be non-intrusively tacked onto an existing object model".

I also focussed less on pure JSON-like objects, and tried to be Pythonic, by adding support for sets, and by allowing non-string keys for dicts.

Currently, objects that get deleted from the model will live on the undo stack as-is. They are not serialized at all. That is fine for a live undo system, but not when either the undo stack needs to be serialized to disk, or when it is to be used for a quick-save feature as described by Raph.

Serializing can easily be done with attrs/cattrs, as @anthrotype pointed out. The UndoManager class will need to grow some hooks to make this possible, though.

I've tried to be thorough with my test suite, but the whole things has not been used in a real situation, so we'll have to see how well it performs.

As a larger thought/question: so far I've mostly been thinking about using an undo manager object for relatively isolated bits of data, for example, "a glyph", "font info", "kerning". But I'm now thinking of into the direction of using one undo manager for a whole object tree, and somehow from the UI allow to undo the topmost undo action for what is currently active/visible. So, for example, even if glyph "a" was changed last, if we are now editing glyph "b", we would only see undo for that glyph. The undo stack is then no longer properly a stack, as actions can be reversed in a different order than they were initially applied. (This has some implications for addition/deletion of objects, but I think it can be figured out what can be done safely.)

Merry Christmas everybody!

davelab6:
@justvanrossum awesome! Would you like to make a quick update to the md doc on this pr referencing your project?

@adrientetar what's needed to merge this?

Happy holidays to all!

justvanrossum:
Yeah, the md definitely needs some updating to reflect the latest discussions. Will do that some time next week.

adrientetar:
> The undo stack is then no longer properly a stack, as actions can be reversed in a different order than they were initially applied.

But then stuff like canUndo() is no longer constant time i.e. you have to start looking inside the list of changes to even know what's undoable in the current context.

Or every object in the tree stores undo changes relevant to itself. e.g. if you have a changeset where glyph width changes and a layer path changes then glyphs stores its change and layer stores its change. Note if you do this you loose total order between changes (as opposed to a single undo stack in glyph) and you probably need to keep track of which change belongs to which changelist (e.g. timestamp and a number to enumerate the changes as they were in the original changelist).

I also think non-linear undo history could be confusing to the user. Suppose I change the width of the glyph then I make edits to a layer (possibly many edits over a long time). Then I switch to another layer, and I hit undo. Changes I made to the other layer don't apply so the undo would probably revert my glyph width change from long ago (note how we need to order and group changes together here, so to know what to revert). Isn't that confusing? Generally things that software does for a user should be clear & easy to understand.

I'm inclined to skip this at least for now. Note there are [papers](https://www.google.com/search?q=non-linear+undo) on non linear undo or undo in a collaborative environment.

adrientetar:
My notes on the delta based undo: https://gist.github.com/adrientetar/aff333972a927c3d0a2c641c27ad605a

davelab6:
I had [a call](https://gist.github.com/davelab6/0e5adcf4092bd778cd2b83f47cba6dd2) with @adrientetar @typefactorynet @raphlinus @justvanrossum just now to discuss this; looks like this doc is ready for merging and @blackfoundry will soon make a PR with their 'snapshot based' undo system, and then look to implement the 'data delta' based approach, perhaps first for glyph objects :)

BlackFoundry:
PR done on TruFont and Tfont repositories.
Could you please review @justvanrossum before @adrientetar can accept and merge?
many thanks

justvanrossum:
Done: https://github.com/trufont/trufont/pull/628#issuecomment-459710626 Btw the other relevant PR is trufont/tfont#18. Cross-referencing FTW :)

adrientetar:
Endgame undo (more or less): https://github.com/adrientetar/Fonte/commit/f94264b51cb1ba728b38f2d6172bae67a633b68c

PR #628, opened 2019-01-23
Wx

BlackFoundry:
There is conflict with @madig changes to __main__.py but we do the same (loading tfont files) with click module, and more debug options.


justvanrossum:
(The title of this PR could be a bit more descriptive.)

Briefly recapping the Blackfoundry approach: There are decorators to wrap UI methods, so the UI code needs to deal with undo/redo as little as possible. Layer objects are (partially) snapshotted for restoring later. For every action, there is a snapshot made before the action, and one after the action. The layer snapshots are partial: if only anchors will be changed, the snapshot will only contain anchors.

This factorisation is nice, as it nicely separates UI code from undo code. However, the decorators are quite complex, and a little hard to read/understand sometimes. The amount of logging is a little excessive, and often distracts from the actual code.

It seems a little wasteful that two snapshots are needed (one before, one after the change), but that may be an unavoidable consequence of the fact that the snapshots are partial. I don't really have a solution to propose, but a sequence of outline edits will result in twice the amount of snapshot data than is strictly necessary.

undoManager.py contains a lot of layer-specific code that should probable live in a different module.

In trufont/tfont#18, the tfont layer object got some undo logic with the `beginUndo()` and `endUndo()` methods. I think that with a snapshot-based approach, all the model object should be responsible for is making snapshots and restoring from snapshots. If the glyph model object should keep a reference to undo data, then that should be an opaque (to the layer object) attribute. If undo is to be used by user scripts, then I think it's better to present scripters with wrapped/proxied objects.

I tried to see if the Blackfoundry approach would be compatible with a more data oriented undo strategy, where changes are recorded instead of snapshots of edited objects. The undo manager infrastructure is definitely capable, as its Action objects are agnostic: they just store callables that can perform anything. That said, the majority of the code in this PR is dealing with specifics such as layer objects.

davelab6:
@adrientetar what is needed to accept/reject this PR? :)

madig:
I think it would help to review this if things were split up into reviewable smaller chunks. This PR sadly looks like a stream-of-consciousness commit log...

davelab6:
@madig thanks for that. I agree. @JeremieHornus ?

BlackFoundry:
I agree the commits log is not very informative.
We might have worked differently to make it more legible to  others.

And thank you for the 3 steps explanation on how to fix this afterwards.
I also like the recommendation, and this is how we worked actually: on a branch named 'wx-bf'. We just merged this branch with the freshly updated master right before doing the PR so, we are in line here.
We just thought it would make more sense to merge masters together for a PR, but it is not necessary actually.

But I don't think the commits log is what should be looked at (they were mainly used as backups): the PR could be seen as one single commit too: 'Undo System' and the differences can be visualized with a  'git diff' before and after the PR:
`git diff 348e9d7264a3b2b2743fc94c59bfdc135f1f0708 5e73f21e39abdafce475652412383138e0d0ef5c > WxTruFontUndoSystem.txt`

Thanks for the link to the article too.

The command line issue thing was for fixing conflict when trying to merge: we had to make a choice.

madig:
The problem is that such a massive PR is very hard to properly review because you have related and unrelated building blocks in one giant pile that you have to understand all at once (which no-one does without spending a day or two trying to work through it). In other projects, PRs usually introduce e.g. tests and base building blocks first and then use them on each subsystem. First you see how stuff is supposed to be used in the tests, then you inspect the building blocks and then you see how they are actually implemented in the subsystems. The commit messages can help with understanding, by explaining why some things were done the way there were done.

It's true that none of this will make this PR have less code, but it will help with getting through it better and more quickly 👍

Issue #623, opened 2019-01-02
ModuleNotFoundError: No module named 'fs'


probonopd:
Getting `ModuleNotFoundError: No module named 'fs'`.
Probably it should be listed in https://github.com/trufont/trufont/blob/master/requirements.txt.

Reference:
https://github.com/linuxdeploy/linuxdeploy-plugin-conda/issues/11

doums:
same issue here, how I can fix this ? (I'm a noob in python)

chrissimpkins:
Workaround until fixed is 

```
pip3 install fs
```

then run again.

probonopd:
Probably `fs` should be listed in https://github.com/trufont/trufont/blob/master/requirements.txt then?

madig:
The requirements are very out of date. Technically you have to install `fontTools[ufo]` to get the dependency.

santhoshtr:
#636 resolved this.

PR #633, opened 2019-02-21
add fs requirement; update fonttools version

jtanadi:
In response to #623


jtanadi:
Didn't realize most up-to-date branch is wx now.

PR #636, opened 2019-03-11
Update requirements to use latest dependencies

santhoshtr:
I know the main work is being done at wx branch. But to make the master branch usable, it is must to udpate the dependencies. 

Addresses #623 too.

This makes the master version usable by cloning the repo, `pip install -r requirements.txt` and then `pip install -e .`


santhoshtr:
it seems dev-requirements.txt is also outdated.

PR #664, opened 2019-08-28
Housekeeping

madig:


madig:
Blargh. Todo: test this in fresh ubuntu1804 image. Something makes Qt's xcb backend not load.

madig:
http://python.6.x6.nabble.com/Travis-and-PyQt-5-12-xcb-plugin-can-t-be-loaded-td5248428.html

Aha!

Issue #577, opened 2018-08-07
OS2CodePageRanges: Thai to Korean Johab bits off by one


madig:
Selecting Thai to Korean Johab writes 
```
    <key>openTypeOS2CodePageRanges</key>
    <array>
      <integer>15</integer>
      <integer>16</integer>
      <integer>17</integer>
      <integer>18</integer>
      <integer>19</integer>
      <integer>20</integer>
    </array>
```
to the UFO. All bits should be plus one.

madig:
Fixed by #674.

PR #674, opened 2019-10-12
fix bit values for OS2CodePageRanges

felipesanches:
(issue #577)

### DIFF
--- a/Lib/trufont/windows/fontInfoWindow.py
+++ b/Lib/trufont/windows/fontInfoWindow.py
@@ -1306,7 +1306,7 @@ class OpenTypeTab(TabWidget):
             elif index > 14:
                 bit += 14
             elif index > 8:
-                bit += 6
+                bit += 7
             if codePageRanges is not None and bit in codePageRanges:
                 item.setCheckState(Qt.Checked)
             else:


felipesanches:
@madig, could you please review & merge this?

madig:
If you make a UFO before this change, select from Vietnamese to Macintosh Character Set and save, apply the change and open the UFO again, Thai to Johab exclusively are checked. Checking Johab and saving does not change the bits saved to the UFO. So there's something else that needs to be fixed.

felipesanches:
thanks, I'll take a look

felipesanches:
it may be yet another off-by-1 bug

madig:
I'm looking at how to add a test for this... custom fields are a pain. So is the structure of the info window...

adrientetar:
You only changed the loading code, not the saving one. Same code in the same file, should probably be factored into a function.

madig:
Just stumbled over that. Made a test, works now. The entire info window could do with a massive refactor... ideally with the GUI construction done in a .ui file and code just for wiring it up...

Now to push it to this branch.

madig:
Done. Thanks again Felipe!

PR #682, opened 2019-10-31
Fix UFO file opening on macOS

### LINKED ISSUE #681: Can't load .ufo file(folders) on macOS
.ufo shows up as a folder on macOS. However, the `FileMode` for `QFileDialog` is set to `QFileDialog.ExistingFile` which does not allow loading folders. (`trufont/Lib/trufont/objects/application.py line 444)

Using the default (`QFileDialog.AnyFile`) file mode will fix this issue. However this puts the onus on the user to open the correct file format.
I would be happy to submit a pull request if needed.

hskim08:
Closes #681


hskim08:
The Info.plist for TruFont needs to notify the OS that .ufo is a package.

PR #693, opened 2020-02-12
Add Show Coordinates submenu

exer2k:
Added functionality providing the drawing of point and Bézier handle coordinates.

The options are available from View menu, from a separate submenu, for more granular control.


madig:
Thanks for sending that in, will take a look over the weekend. Please format the code with black in the meantime :)

exer2k:
Whoops! I wasn't aware there are formatting rules applied to the code. Will fix that in a minute.

madig:
I use https://github.com/pre-commit/pre-commit to run checks before committing.

exer2k:
This is a discussion for another issue, but should we add the formatting and/or pre-commit check instructions to `README`?

madig:
Yes :) Can look into it later.

madig:
(Small note on the .gitignore file: you can define git ignore patterns on your local machine by modifying `.config/git/ignore` (maybe you need to set the `core.excludesfile` parameter), so you don't have to touch the `.gitignore` file in random projects you contribute to.)

madig:
Hm, with a fresh configuration, all the options are toggled in the view menu but not necessarily active. I think you want them off by default?

exer2k:
You're right, I'll take a look at it.

madig:
Ok, looks good to me. Thanks! :)

PR #697, opened 2020-03-01
add 'type1' to fonttools extras requirements as it's needed by ufo-extractor

anthrotype:
without this, pip installing in editable mode on macOS fails with

```
pkg_resources.DistributionNotFound: The 'xattr; sys_platform == darwin and extra == type1' distribution was not found and is required by fonttools
```


anthrotype:
correction: no error occurs while doing `pip install -e .`; it's only when attempting to run the `trufont` binary that that error occurs.

unfortunately pip is not smart enough to automatically augment the set of extras required if any dependency requires a different set of fonttools extras, so we need to do that ourselves at the top level.

PR #710, opened 2020-04-10
Fix open/import file dialog filter (fixes #705)

### LINKED ISSUE #705: [0.6.3] "All supported files" in Open/Import dialog
Hi!

The current open/import dialog has a single filter labelled "All supported files". So I need to read code to find out what files are supported:
https://github.com/trufont/trufont/blob/a5d0282420b44268e45c1e12fa9cfe4e63c69b29/Lib/trufont/objects/application.py#L403-L414

With regard to UX, that's a bit unfortunate. It would be great to have one filter per type so that as a user I can open the dropdown and get a sense of supported types.

PS: The split between open and import seems artificial from a UX point of view to me. Maybe these two dialogs can be merged?

Thanks and best, Sebastian

hartwork:
Fixes #705

For your consideration. Review on commit level advised.

### DIFF
--- a/Lib/trufont/objects/application.py
+++ b/Lib/trufont/objects/application.py
@@ -412,13 +412,11 @@ class Application(QApplication):
                     ]
                 )
                 supportedFiles += "*.otf *.pfa *.pfb *.ttf *.ttx *.woff"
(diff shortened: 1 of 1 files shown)


madig:
Thanks!

PR #719, opened 2020-05-12
set index of combo box during initialization

### LINKED ISSUE #718: PointSize index mismatch in Metric window during initialization
When first opening the metric window the initial point size is read from the line view, setting the initial point size to 200. However, the combo box index is still 0. (See attached picture)

Selecting the combo box and pressing down then up will cause the point size to change from 200 to 75, then to 50.

A fix is to set the combo box index during initialization.

Screen Shot 2020-05-12 at 12 03 06 PM

hskim08:
Closes #718


madig:
Thanks!

Issue #555, opened 2018-04-21
Windows/macOS: task bar integration wonky


madig:
On Windows, launching Trufont and hovering over the task bar symbol says that it's called "IDLE 3.x...", you also can't pin the app and the app icon won't show.

On macOS, the application is called "Python".

hskim08:
This seems to work for macOS: https://stackoverflow.com/a/54755290.
But it requires pyobjc-framework-Cocoa for python3.
I'm not familiar with how to handle pip installs for platform specific cases.
I tested it on my own branch: https://github.com/hskim08/trufont/tree/fix-app-title

PR #720, opened 2020-05-13
Set application title on macOS

hskim08:
Changes the application title from the default "Python" to "TruFont" for macOS.
It requires pyobjc-framework-Cocoa.

Fixes the macOS side of #555


madig:
Please add the dependency to setup.cfg with a platform marker. The comment can then be removed.

madig:
Thanks!

Issue #438, opened 2017-05-10
Names of glyphs in glyph view cut off slightly.


madig:
![bildschirmfoto vom 2017-05-10 23-02-01](https://cloud.githubusercontent.com/assets/380829/25921160/bf4dbef0-35d4-11e7-98b6-daf890bca2e7.png)

The cell containing the name of the glyph should be made slightly taller.

driehuis:
Bumping GlyphCellHeaderHeight to 15 (from 13) resolves the clipping for me.

PR #764, opened 2021-04-09
Bump GlyphCellHeaderHeight to 15 (from 13).

driehuis:
Bump GlyphCellHeaderHeight to 15 (from 13). This resolves the clipping mentioned in issue #438, at least on my system (Ubuntu 20.04LTS).

### DIFF
--- a/Lib/defconQt/representationFactories/glyphCellFactory.py
+++ b/Lib/defconQt/representationFactories/glyphCellFactory.py
@@ -4,7 +4,7 @@ from PyQt5.QtGui import QColor, QFontMetrics, QPainter, QPainterPath, QPixmap
 from defconQt.tools import platformSpecific
 from defconQt.tools.drawing import colorToQColor
 
-GlyphCellHeaderHeight = 15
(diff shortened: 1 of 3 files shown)


driehuis:
Actually, this improves the situation but does not fix it completely. I'd still apply this change though. A thorough fix requires rendering a few characters and measuring their height in pixels, which is beyond my knowledge of PyQt.

adrientetar:
Thanks; could you bump it only for linux maybe? So expose a constant in platformSpecific that goes like 

```py
if sys.platform.startswith("linux"):
    return 15
return 13
```

driehuis:
Sigh. My update to this PR seems to have dragged in other changes. Rather than try to fix my git-fu I'll close this PR and create a new one.

adrientetar:
You can do git rebase -i HEAD^^^^ and then git push -f

driehuis:
I've been bitten by `git rebase` and `git push -f` in the past (admittedly after trying to fix a more complex merge fufu), but I vowed there and then to just redo the branch if I can easily afford to, rather than risk shooting myself in the foot.  :-)

PR #792, opened 2021-10-03
Bump GlyphCellHeaderHeight to 15 (from 13) for Linux.

driehuis:
Bump GlyphCellHeaderHeight to 15 (from 13) for Linux. This fixes the clipping mentioned in issue #438, at least on my system (Ubuntu 20.04).

Reworked to use platform specific helper function, replacing PR 764. For optimum results, should be reworked to actually measure UI font height.

### DIFF
--- a/Lib/defconQt/representationFactories/glyphCellFactory.py
+++ b/Lib/defconQt/representationFactories/glyphCellFactory.py
@@ -4,7 +4,7 @@ from PyQt5.QtGui import QColor, QFontMetrics, QPainter, QPainterPath, QPixmap
 from defconQt.tools import platformSpecific
 from defconQt.tools.drawing import colorToQColor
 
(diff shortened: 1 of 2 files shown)


adrientetar:
Thanks!

driehuis:
I may actually have come up with a proper fix for this issue, but I cannot test it across platforms. Adding this at line 95 of Lib/defconQt/representationFactories/glyphCellFactory.py seems to do the right thing:

```
       fontMetrics = QFontMetrics(headerFont)
       GlyphCellHeaderHeight = fontMetrics.height()
```

It seems to be called only once, so no performance impact AFAICT, but I'm also not convinced this is the cleanest place to perform this magic.

driehuis:
Hmmm, scratch that. The fix isn't consistent. I'll keep looking for a proper fix that takes glyph height into consideration.

